PR #343, opened 2014-06-10
Improve regions and date handling

abryzak:
Regions are currently drawn on top of grids, which means if there are any grid lines with labels, those labels are completely obscured by the regions when the opacity is 1.

The date update separates how dates in the data and on the axis labels are handled, allowing UTC dates in the data, which can be manipulated correctly without daylight savings time causing issues.


Mshardul:
I am not sure, if I am doing something wrong or what; but this problems still exists. 
I am using v0.7.13 (and the problem was there in v0.7.12 as well).

> .c3-event-rects{
>         fill-opacity: 1!important;
>         fill: red;
>     }

would not show the grid-lines.

PR #555, opened 2014-09-11
phantomjs based exporter for c3 charting library

yuvii:
a configurable command line script based on phantom.js that can export charts to files (server-side of course). Supports `.png`, `.gif`, `.jpeg` and `.pdf` formats, with a config file that allows for extra js (for c3 extentions and over-rides) and extra css. Basic usage:

```
path/to/phantomjs outpufile.png options [WxH]
```
- output can be any path with any of the above supported formats
- c3options is a JSON string to be used in `c3.generate( options )`

The config file also includes a `template` which can be over-ridden. I will later work on an example page showing the usage


umesh-naik:
Hi yuvii, I have got a problem. I use Linux and the issue I'm facing is, when I take print of C3 charts, the charts are getting blurred. And since there is no static built for Linux version, I'm using 1.9.8 v. But when I passed same html code for generating pdf using 2.0.0 v in Windows, charts are clear. Can you please explain or have any solution for this?

PR #603, opened 2014-10-07
configurability of domain boundaries for zoom

### LINKED ISSUE #570: Configurable domain boundaries
I think there should be a option to extend domain boundaries. (example usage is reloading data from future/history using onzoomend callback from #569)

michalkop93:
add ability to set domain boundaries. closes #570


masayuki0812:
Sorry for the late response. I merged this PR and fixed to zoom out based on the new options `zoom.x.min/max`. I think it makes zooming behaviour more intuitive. Thanks.

PR #641, opened 2014-10-22
Adding named array support for order

chris-teague:


masayuki0812:
Please let me close at the moment. Thanks.

PR #747, opened 2014-11-21
readded ability to set axis.x.values to function

michalkop93:
Again allows to set generator function to axis.x.values for dynamic and precise axis


masayuki0812:
Sorry for the late response. Merged and thank you for working on this.

PR #813, opened 2014-12-08
add option textAnchor: 'start' to put text near the start (instead of th...

dungsaga:
...(intead of the end) of grid line

Example: http://jsfiddle.net/thto4uqp/


masayuki0812:
This feature is already implemented in `0.4.10`, so let me close. Thanks.
https://github.com/masayuki0812/c3/releases/tag/0.4.10

PR #833, opened 2014-12-17
Introduce tooltip.position callback function

ku:
Added a callback function to control the position of the tooltip programmatically to tooltip configuration.
`circle` element is given if `grouped` option is false and `region` element is given if it's true.

This callback function allows to show tooltip on the top of the data points like this.

with following configuration.

```
        tooltip: {
          grouped: false,
          position: function (data, width, height, element) {
            var chartOffsetX = document.querySelector("#chart").getBoundingClientRect().left;
            var graphOffsetX = document.querySelector("#chart g.c3-axis-y").getBoundingClientRect().right;
            var tooltipWidth = document.getElementById('tooltip').parentNode.clientWidth;
            var x = (parseInt(element.getAttribute('cx')) ) + graphOffsetX - chartOffsetX - Math.floor(tooltipWidth/2);
            var y = element.getAttribute('cy');
            var y = y - height - 14;
            return {top: y, left: x}
          },
          contents: function (d, defaultTitleFormat, defaultValueFormat, color) {
            d = Math.floor(d[0].value)
            return '<div id="tooltip" class="module-triangle-bottom">' + d + ' wows</div>'
          }
        }
```

### DIFF
--- a/spec/tooltip-spec.js
+++ b/spec/tooltip-spec.js
@@ -1,29 +1,38 @@
 var describe = window.describe,
     expect = window.expect,
     it = window.it,
+    jasmine = window.jasmine,
+    beforeAll = window.beforeAll,
     beforeEach = window.beforeEach;
 
 describe('c3 chart tooltip', function () {
     'use strict';
 
     var chart, d3;
+    var tooltipConfiguration;
 
-    var args = {
-        data: {
-            columns: [
-                ['data1', 30, 200, 100, 400, 150, 250],
-                ['data2', 50, 20, 10, 40, 15, 25],
-                ['data3', 150, 120, 110, 140, 115, 125]
-            ]
-        }
+    var args = function () {
+        return {
+            data: {
+                columns: [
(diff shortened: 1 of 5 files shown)


aendrew:
Oh rad! Good to know it can do this now. The documentation definitely needs to be updated to include this!

parisnakitakejser:
Thanks a lot, you save my night! :+1:

stuartm:
Is it possible to modify top while leaving left unchanged, or vice-versa, or is it always necessary to recalculate both?

vinikento:
There's just one problem:
When values match on the graphic, it only show one data.

PeddiMahesh:
If we are using xs in data how to get 'cx' and 'cy' values of point.

opensrcken:
This does not work for bar charts.

jayantsr83:
How this code will work with donut chart? Any Help?

JRPendarvis:
Can the function be an external file?

PR #860, opened 2015-01-03
Calling resize on each and every window resize drammatically slows down the browser

### LINKED ISSUE #857: Calling resize on each and every window resize drammatically slows down the browser
If you have a lot of charts and use Firefox (this issue appears in other browsers as well, but it slows down especially Firefox), resizing the window can be a huge pain (resizing is fired not only when you resize the window itself, but also when you open Firebug, for example). This happens because `$$.api.flush();` is called on every resize event call. Which can be tens of times. Instead of calling it on each resize event call, I suggest using the following code:

```js
        if (window.onresize.add) {
            var timeout;
            window.onresize.add(function () {
                config.onresize.call($$);
            });
            if (config.auto_resize) {
                window.onresize.add(function () {
                    if (timeout !== undefined) {
                        clearTimeout(timeout);
                    }
                    timeout = setTimeout(function() {
                        $$.api.flush();
                    }, 100);
                });
            }
            window.onresize.add(function () {
                config.onresized.call($$);
            });
        }
```

`$$.api.flush()` will be called not more than once per 100ms (the number is random here, I didn't make a lot of tests, which one is the best).

Also, please note `auto_resize` setting. It could be used to turn off auto resizing the charts. By default it should be true for backwards code compatibility.

I hope, you got my idea and this can go to the core.

denisvmedia:
This commit resolves #857

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -7,6 +7,7 @@ c3_chart_internal_fn.getDefaultConfig = function () {
         padding_right: undefined,
         padding_top: undefined,
         padding_bottom: undefined,
+        auto_resize: true,
         zoom_enabled: false,
         zoom_extent: undefined,
         zoom_privileged: false,
--- a/src/config.js
+++ b/src/config.js
@@ -7,7 +7,7 @@ c3_chart_internal_fn.getDefaultConfig = function () {
         padding_right: undefined,
         padding_top: undefined,
         padding_bottom: undefined,
-        auto_resize: true,
+        resize_auto: true,
         zoom_enabled: false,
         zoom_extent: undefined,
         zoom_privileged: false,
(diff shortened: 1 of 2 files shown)


masayuki0812:
Hi,
Thank you very much for this fix!! Your code looks nice, but one thing, can you change `auto.resize` to `resize.auto`? It looks more c3-way for me.

denisvmedia:
I updated the name (the first commit in the update has a wrong comment though, but internally it's correct; I'm sorry for that).

aendrew:
Hi @denisvmedia! Could you please squash your commits?

`git rebase -i HEAD~3` -- [more info](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

ajainarayanan:
Is this change still going in by any chance? This improves the performance by so much.

masayuki0812:
@denisvmedia Sorry for the late response. Thanks!!

PR #879, opened 2015-01-09
Control whether tooltip is displayed by value

chris-vaszauskas:
I want to be able conditionally control whether a tooltip is displayed or not for a given value. If I return undefined from a custom `valueFormat` function, I don't want the tooltip for that value to be displayed.


aendrew:
Great idea! PR looks good to me; @masayuki0812, care to merge? :+1:

masayuki0812:
Looks good. Thanks!

Issue #895, opened 2015-01-15
Axes should have a "position" argument


aendrew:
As per the discussion in #759, the `y2` axis is meant to be less constrained than `y`, but using `y2` is the only way of getting that axis on the right-hand side. As such, as a `position` option for each axis would be helpful.

aendrew:
I'm working on a pull request for this, but need a bit of guidance. How should position be treated if the axes are rotated? Should position just be ignored? Translated (I.e., "left" becomes "bottom", "right" becomes "top")?

masayuki0812:
@aendrew Thanks for working on this. I think it should be ignored at this moment because it seems a little bit confusing and there would be not so much use case to translate (rotate dynamically).

aendrew:
@masayuki0812 Sounds good! I think my PR won't merge at this point anyhow, so I'll just close that.

Issue #112, opened 2014-04-02
More than 2 Y axes


nathancole:
Any chance at support for more than 2 Y axes? :-)

gnomeontherun:
+1 This would be useful!

gfoiani:
+1

TwentyOneSolutions:
+1

jiangmf:
+1 would love this feature

nanoix9:
Is this supported now?

filipkral:
+1
We need more than 2 y axes very often.

ultravail:
+1
would be very helpful

MiguelGiordani:
+1
I am migrating graphs from Jpgraph (PHP) and I didn't find any free project to create a graph with multiple axis like the old lib.... Waiting for c3..........

petgus:
+1
This is a deal breaker for me.
This lib which also uses d3 seems to manage it. Using that until c3 has this feature.
[](https://plot.ly/javascript/multiple-axes/)

Boorj:
+1 it's a quite common situation.. waiting for c3js, need to recreate this
![2016-06-26_01-52-18](https://cloud.githubusercontent.com/assets/2145313/16358776/19ad8d08-3b49-11e6-8804-2ad79f60046e.png)

Thinking about adding marker + additional tick info. Or even custom svg lines...

Michael-AUT:
+1 for me too

kunKun-tx:
+1 
I need to have an additional axis for label purpose.
I ended up using vanilla D3 for this. http://jsfiddle.net/Ea4mT/1/

gneatgeek:
I see a lot of +1s here and am curious if there are any plans to do this for C3?  Also 👍, very useful

jasonhrv:
+1

jpruetting:
+1

mipapo:
+1

neal-morgansolar:
+1

ljubomirb:
+1

utkarsh8085:
+1
Guys, please implement this feature. We all needed this very badly.
Thanks in advance.

aendrew:
Locking this issue because I'm really sick of people +1'ing it instead of using the emoji or simply subscribing to it.

PR #899, opened 2015-01-15
Adding position argument for Y axis (IN PROGRESS, PLEASE DON'T MERGE YET)

aendrew:
I've taken a brief stab at adding a position argument for #895, but it still needs work. Will update this PR as I make progress.

Currently it allows you to specify position for Y (`axis.y.position` == "left" || "right"). What it doesn't do is size everything properly if there's already an axis on that side (I.e., you show both `y` and `y2` while `y` is set to "right"), or have any clue how to deal with rotated axes.

## TODO before ready to merge

1. Add support for Y2
2. Improve unit tests
3. Figure out how to deal with rotated axes
4. Add handling for when there are two axes on the same side.
5. Fix tooltips.

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -113,8 +113,8 @@
         $$.defocusedTargetIds = [];
 
         $$.xOrient = config.axis_rotated ? "left" : "bottom";
-        $$.yOrient = config.axis_rotated ? (config.axis_y_inner ? "top" : "bottom") : (config.axis_y_inner ? "right" : "left");
-        $$.y2Orient = config.axis_rotated ? (config.axis_y2_inner ? "bottom" : "top") : (config.axis_y2_inner ? "left" : "right");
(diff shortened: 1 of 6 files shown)


aendrew:
Note also that I haven't really added anything in terms of internal functions and have just tried to deal with the extra config option via conditionals; TBH, it might be worthwhile creating some sort of intermediating function to prevent some of the long ternary operations (core.js, ln. 111-112, for instance). I just don't know the internal API well enough to know whether I'm already duplicating something.

Considering there's some discussion about redoing the axis API in #112, I almost wonder whether adding this at the moment is worthwhile — if axes were modular enough to add an arbitrary number, much of the positioning stuff I'm struggling with would already be taken care of.

aendrew:
Oh, and it appears I've somehow botched tooltips as well...

aendrew:
Closing because this is too confusing right now and will probably disrupt something else.

PR #904, opened 2015-01-16
Update c3.axis.js

EvanGranthamBrown:
On axis labels, allow lines to break on hyphens and slashes as well as spaces.

### DIFF
--- a/src/c3.axis.js
+++ b/src/c3.axis.js
@@ -110,7 +110,7 @@ function c3_axis(d3, params) {
             // this should be called only when category axis
             function splitTickText(d, maxWidth) {
                 var tickText = textFormatted(d),
-                    subtext, spaceIndex, textWidth, splitted = [];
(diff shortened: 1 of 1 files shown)


aendrew:
I'm not sure this is desirable — what happens with dates, for instance?

EvanGranthamBrown:
Hmm, good point about dates - that hadn't occurred to me. Negative numbers might conceivably be a problem, too.

The main issue is that any time you join two longish words with a slash or a hyphen, the result is apt to be too long for the space available, resulting in a line break mid-word. One could of course work around this by putting a space after the slash/hyphen, but then if you *don't* get a line break there, you have a space just hanging out and it looks ugly.

I'll give it some thought and see if I can come up with a better solution. Thanks!

masayuki0812:
Thanks for the PR, but please let me close at the moment. Thanks.

Issue #926, opened 2015-01-21
Repeatedly calling c3.generate leaks memory in 0.4.9


ergo:
Hello,

I got reports that pages that constantly regenerate our chart objects use up too much memory and kill browser tabs:

http://plnkr.co/edit/l0JOvGBWRDEv0b3fwywr?p=preview - here is an example with 0.4.9 version

http://plnkr.co/edit/wEQiMgBeBzVAwYTdijKD?p=preview - here is an example with 0.4.8, it seems to leak way less and is way more stable according to my tests.

aendrew:
@ergo Interesting — is there any reason you're calling `destroy` and then loading more data instead of just using `api.load`?

ergo:
Yes, the reason is c3 doesn't support setting some options on load() - like changing chart types of existing series or setting new labels, or setting x tick values - this only works when generate() is called.
So my workaround is to just generate whole chart - not ideal but seems the only option right now when I will have generative charts and the values might change on update.

And yes I've tried downgrading to 3.5.0 and 3.4.3 - you can fork and edit the plunkers the 0.4.9 one will quickly reach 1.5gb mem regardless of d3 version used :(

aendrew:
@ergo Ah, I see — seemed like the data remained sort of constant, which was why I was wondering about `load()` (FWIW, I do the exact same thing with times/axisJS whenever the user changes chart types, would be really nice to be able to modify charts without having to call `c3.generate()` again. Might be worth creating a separate issue about improving `load()`'s functionality).

But yeah, tab just crashed again. Yikes, that's definitely an issue. Marking as bug.

aendrew:
There is definitely something leaking memory. I just tried to do a screencast using axisJS and it crashed the browser tab after about 2 minutes on my recent-model Macbook Pro. I'm going to investigate this today, this bug seems pretty dire...

aendrew:
I'll update this comment as I find more. In the meantime:
+ [Plunker](http://plnkr.co/edit/tWiBxCxPq2URD7Vpdwet)
+ Test code: ```
  var generateData = function() {
    var data = [];
    for (var i = 0; i < 5; i++) {
      var datum = [];
      datum.push('dataset ' + i);
      for (var j = 0; j < 20; j++) {
        datum.push(Math.random() * 1000);
      }
      data.push(datum);
    }

    return data;
  };

  var int;

  var toggleTest = function() {
    if (!int) {
      int = setInterval(function(){
        c3.generate({
          data: {
[...]

PR #938, opened 2015-01-24
Adds c3_chart_internal_fn.isTabVisible to disable transitions on inactive tabs. Fixes #924.

### LINKED ISSUE #924: Chart does not render properly when it was generated in not active tab in chrome
I opened http://c3js.org/samples/chart_combination.html from http://c3js.org/examples.html by clicking at link and opening it in new tab. Then I waited few seconds and went to combination example. It was not rendered (Actually it was but wrong).  
To get chart visible I had to resize window. If it is only one tab in chrome, this bug does not seem to appear.

Google Chrome 39.0.2171.99 m

aendrew:
This adds a new function, `c3_chart_internal_fn.isTabVisible`, which uses the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API) to return whether the page is inactive or not. I've then called it `c3_chart_internal_fn.redraw()` on line 614 of `src/core.js`, which short-circuits the conditional and renders the chart without using `d3.transition` (which uses the `requestAnimationFrame` API, which doesn't work on inactive tabs) if the chart's browser tab is inactive.

See #924.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -611,7 +611,7 @@ c3_chart_internal_fn.redraw = function (options, transitions) {
         });
     }
 
-    if (duration) {
+    if (duration && $$.isTabVisible()) { // Only use transition if tab visible. See #938.
         // transition should be derived from one transition
         d3.transition().duration(duration).each(function () {
             var transitionsToWait = [];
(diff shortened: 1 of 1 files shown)


masayuki0812:
Looks nice. Thanks!

aendrew:
Yay! Thanks! :+1: :shipit:

ergo:
I have serious doubts if this is a proper fix for the issue, after all it should not leak and GC correctly regardless of visibility?

aendrew:
@ergo The GC and memory leak issue is a separate one; I totally agree this is not the fix for that. This is merely to resolve the issue with requestAnimationFrame not firing in inactive tabs, resulting in the chart not rendering if `c3.generate()` is called in an inactive tab.

masayuki0812:
@ergo Yeah, this does not fix the #926 , but seems making the speed of leak slower like 0.4.8 as you reported.

PR #990, opened 2015-02-10
Correct input for generateGetAreaPoints

### LINKED ISSUE #980: Data labels for multi-series area chart have incorrect position
If a chart is configured with more than one area (or area-spline or area-step) series, all but the first series has data labels placed incorrectly. Here's an example: http://jsfiddle.net/fa5t4ba8/ 

When combined with flow, probably because it doesn't redraw the labels, there is an error that appears in the JS log: Uncaught TypeError: Cannot read property 'value' of undefined. Example here: http://jsfiddle.net/Lxxh185v/

rhenna:
Fixes an issue where generateGetAreaPoints was taking in the wrong input
causing labels for area charts to be incorrectly positioned. Fixes #980.

### DIFF
--- a/src/text.js
+++ b/src/text.js
@@ -62,7 +62,7 @@ c3_chart_internal_fn.getTextRect = function (text, cls) {
 };
 c3_chart_internal_fn.generateXYForText = function (areaIndices, barIndices, lineIndices, forX) {
     var $$ = this,
-        getAreaPoints = $$.generateGetAreaPoints(barIndices, false),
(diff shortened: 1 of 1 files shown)


aendrew:
Confirm fixes #980.

Before: http://jsfiddle.net/fa5t4ba8/
After: http://jsfiddle.net/aendrew/s943qop1/1/

masayuki0812:
As I wrote in #980, I fixed this issue in the same way. Thank you for your PR.

PR #1014, opened 2015-02-20
Chrome Glitch

GordonSmith:
Setting a class on the "body" (even for a brief moment), is causing chrome to
change other non C3 elements positions.

Signed-off-by: Gordon Smith <gordon.smith@lexisnexis.com>


masayuki0812:
I fixed this issue by myself. I think now it's ok and I'll release this fix in the next version 0.4.10 shortly. Thanks.

Issue #1031, opened 2015-02-26
Add optional headers to charts


davidpoore:
We've already written some code to handle this in the Pivotal Tracker fork of c3 (https://github.com/pivotaltracker/c3) but I am doing some personal work to clean up that code and open pull requests for our most desired features so that they can be part of the c3 core library.

For the charts designs I'm working with, I want to have a header bar with a different background color from the rest of the chart, where the title, legend, etc. can live. I think a simple config option for `header` that allows you to specify a header color would be useful, especially in conjunction with adding titles to the library (see issue #1021).

I imagine the config will look something like this:
```
header: {
    show: true,
    color: '#EBEBEB',
    // for showing a border line along the bottom of the header
    border: {
        show: true,
        color: '#000'
    }
}
```

michaelrodov:
Closing
This was addressed via Add headers for charts #1032

PR #1032, opened 2015-02-26
Add headers for charts

davidpoore:
Issue #1031 - add optional config option to add header to a chart


aendrew:
Hi!
What exactly does this do? All I can get it to do is draw a line on top. Am I doing something wrong?

http://jsfiddle.net/aendrew/fnbno2os/2/

davidpoore:
In conjunction with the title, it allows you to add a header line (with or without a bottom border) that draws behind the title. It was cleaner than trying to append the headers we wanted after drawing the initial SVG and allows chart headers that look like this:

screen shot 2015-03-27 at 9 36 58 am

That header at the top, including the title, legend, etc. is now entirely rendered using c3, instead of having to be styled or drawn separately.

davidpoore:
Also, the size of the header is based on the top padding of the chart, so that its size increases automatically if you increase the size of the title or add more padding.

aendrew:
Ahhh... Okay, cool. Will take another look later today.

davidpoore:
@aendrew fixed up merge conflicts and pushed this again - hoping to get all my open pull requests merged in soon!

aendrew:
@davidpoore Cool, thanks! Will try to RTBC it today.

aendrew:
@davidpoore Hi, huge apologies for the utterly atrocious length of time it's taken me to get to this!

I've reviewed it and predominantly LGTM, though a few points:

a. Took me awhile to realise you need to have a value for `padding.top` for the header to appear, and then it was black when it did. Most users won't do more than set `header.show` to `true`, so the defaults should be such that simply doing that yields something in the style of the rest of the chart. The styles you have in [that Dribbble](https://dribbble.com/shots/1879542-Story-type-breakdown-chart?list=users&offset=2) are perfect. 
b. Even then, I'm not sure padding is the best attribute to use in this case, as I'd expect a chart with a header and 10px of top padding to have a header, and 10px additional space between the chart and the header. TBH, people will probably want to tweak what the header looks like less than they'll want to tweak the chart itself, so the header shouldn't interfere with that user goal. Maybe put the header above the padding, and add a height configuration option to the header? Regardless, it should look good and be able to comfortably accommodate the title if only `title.text` and `header.show` have values.
c. Thanks once again for adding a spec! :cookie: 
d. Would love to have a footer option (and the ability to put the title there) given the requirements of the charts I make, but that's probably another PR for another day... :smile: 

Not sure what @masayuki0812 thinks, but it'd be great to get this into 0.4.11 given how well it works in conjunction with the new title feature. 

Thanks for your ongoing efforts, I really appreciate them! I'm ultimately wanting to rip out the awful title creation functionality in [Axis](http://github.com/times/axisJS) and [replace it with the native C3 version](https://github.com/times/axisJS/issues/84) once 0.4.11 drops, so huge thanks for working on this!

otisg:
+1 for having support for the header.... though looks like there are some conflicts that need to be resolved here.

davidpoore:
Sorry it took me so long to get back to this.

1. Header and header border both now have default values so they appear even if the only option passed to them is `show: true`
2. Header height completely decoupled from padding, defaults to 15, and can be set via config option.
3. Adding padding to the top of a chart increases spacing between header and chart.

masayuki0812:
Looks good and this feature is useful in some use cases, but I think this should be a plugin because c3 should be small as much as possible and I don't think this feature seems not required by most of users.
But thanks anyway. I'll merge this feature as a plugin in the future.

pchaganti:
👍

kt3k:
Closing because this is now a plugin candidate. #2034

Issue #1033, opened 2015-02-26
Custom legend item tile size & legend item padding


davidpoore:
For some of our chart designs, we want to have legend element tiles that were lines rather than boxes - i.e. legend element tiles that look like this: 
![screen shot 2015-02-26 at 2 00 55 pm](https://cloud.githubusercontent.com/assets/6174013/6401797/1d8617a4-bdc0-11e4-9a65-1e353d591f87.png)

Also we want to be able to control the amount of space in between each legend element for more fine-grained control over how it is displayed.

So I want to add custom config options for these:
```
legend: {
    // amount of padding to put between each legend element
    padding: 5,
    // define custom height and width for the legend item tile
    item: {
        tile: {
            width: 15,
            height: 2
        }
    }
}
```

The default values for width and height would be the current legend item size.

michaelrodov:
pull request filed
closing

jerrylow:
Note this needs to be added to the documentation.

fabiofcferreira:
I've been searching for ways to customize tiles, e.g. (adding styles using ::after or ::before) and I found this.
I think this should be added to the docs as soon as possible.

Thanks for this awesome library. 😄

taliffsss:
how can i manipulate legend text like i want to split a text. how could i do that thanks and advance

fabiofcferreira:
@taliffsss I guess you could use CSS for that but to be honest, if you can, use D3 as it is the base for C3.js.

mvidailhet:
This useful configuration is not in the doc yet. That's too bad! I just found it thanks to Google.

rogeriolino:
Item padding isn't working when legend is 'inset': 

Inset
https://jsfiddle.net/rogeriolino/by8vwsa5/

Default
https://jsfiddle.net/rogeriolino/by8vwsa5/4/

PR #1034, opened 2015-02-26
#1033 custom legend config

davidpoore:
Allow for custom legend configuration options - users can change height and width of tile, and specify custom amounts of extra padding between legend elements.

I changed the legend tiles from rects to lines because I felt it made the drawing logic a lot easier for me to visualize, and because it made the centering of the tile elements almost automatic.

As a side note - currently, the DOM for the jasmine tests does not seem to be getting cleaned up in between each spec running - this could cause some problems for people writing specs in the future as they may accidentally grab old DOM elements when they try to do selections (which is what seemed to be happening to me). 

However, that is probably a new issue and not related to the actual functionality represented by this pull request. I was able to get around it by changing the names of the `data` sets in the `args`.

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -1092,6 +1092,9 @@
             legend_item_onmouseover: undefined,
             legend_item_onmouseout: undefined,
             legend_equally: false,
+            legend_padding: 0,
+            legend_item_tile_width: 10,
+            legend_item_tile_height: 10,
             // axis
             axis_rotated: false,
             axis_x_show: true,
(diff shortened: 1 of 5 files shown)


mwean:
:+1: I really need this because my legend labels are overlapping

davidpoore:
@aendrew made sure this one was ready to go, too - hoping to get these all merged soon

masayuki0812:
Looks very nice. Thanks! @davidpoore

PR #1043, opened 2015-03-05
Fix Function.prototype.bind polyfill by using Mozillas suggested script

matthelm:
### Problem

This PR fixes a bug with the `Function.prototype.bind` polyfill where arguments passed into `bind` weren't being included when the function was eventually executed. I've included three examples showing the expectation without a polyfill, what's currently happening in master, and how the polyfill should be implemented in order to emulate the baseline.

[No Polyfill (baseline)](http://jsbin.com/jugelaheji/1/edit?js,console)
[C3 master (broken)](http://jsbin.com/sixoqeduco/1/edit?js,console)

### Solution

[C3 matthelm:phantom-polyfill (fixed)](http://jsbin.com/qunesepive/1/edit?js,console)

The code for the PR is copied from MDN's suggested polyfill [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Polyfill). I'm unsure how to best write tests for this code, and because the current polyfill didn't have tests I'm submitting this PR without any. That said, I'm happy to add them.

### DIFF
--- a/src/polyfill.js
+++ b/src/polyfill.js
@@ -1,7 +1,25 @@
-// fix problems using c3 with phantomjs #578
-Function.prototype.bind = Function.prototype.bind || function (thisp) {
-    var fn = this;
-    return function () {
-        return fn.apply(thisp, arguments);
-    };
-};
+// PhantomJS doesn't have support for Function.prototype.bind, which has caused confusion. Use
+// this polyfill to avoid the confusion.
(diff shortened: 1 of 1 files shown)


matthelm:
@masayuki0812 Please review at your convenience.

masayuki0812:
Looks nice. Thank you very much!!

PR #1053, opened 2015-03-09
Fix #1052

### LINKED ISSUE #1052: Pie, Donut & Gauge Chart Hide Serie on Click
An error is thrown when hidding a serie with the "onclick" event on pie, donut and gauge charts. See http://jsfiddle.net/3x78m8sf/ . 

I used the logic of c3's legend onclick
```
onclick: function(d, i) {
    chart.toggle(d.id);
    chart.revert();            
}
```

The error is not thrown on other chart type

GermainBergeron:
Added check as function might return null


masayuki0812:
I could not merge this PR, so I pushed as my commit: https://github.com/masayuki0812/c3/commit/fb5154338348a0adbcdee067543d497c7ad87bde
Thanks!

PR #1069, opened 2015-03-19
Move opacity styling from js to css so that they can be modifiable.

bjlbernal:
We prefer to have the bars and pie charts 75% opaque unless focused upon by the cursor. A slight variation of the default. In order to pull this off with the pie charts we needed to have opacity abstracted out of the arc charts. Out of curiosity I checked the necessity of the bar chart opacity 0 values being set to 0. We didn't find removing them to cause any errors so we are submitting this as a pull request. The default values were just moved out of the js and put in the css.


Funfun:
+1 nice PR.

aendrew:
I'm in support of this, not least because it will make it easier to render as PNG.

@bjlbernal Apologies for the delay in getting to this, would you mind please squashing your commits? Thanks!

bjlbernal:
@aendrew This is like ages old, but I finally got some free time to do this. After squashing and trying to sync up to the current release candidate I wasn't sure I was doing it right so I just created a new branch with the current release candidate and committed my changes here instead. I hope this is can be done. If I need to submit a new PR, let me know.

be910bb32f594bb0025db70f07e0317b3292b1e8

aendrew:
@bjlbernal Apologies for how astoundingly long it's taken to get this merged in — in lieu of having to do an hour of rebasing/merge conflict resolving, I've recreated your PR at #2069 (hilariously, exactly 1000 issues/PRs after you initially submitted it. 😓).

PR #1073, opened 2015-03-20
add expand duration config for pie, gauge, and donut charts

Shrugs:
This PR adds the config

```js
{
    pie: { // or gauge, or donut
        expand: {
            duration: 100
        }
    }
}
```

for controlling the duration of the expand animation. It was too abrupt for my application so I figured I'd propose my changes.

The `expand` config defaults to an object (which still evaluates to `true` in the fn near line 4784), so there are no breaking or major changes introduced.

Cheers


masayuki0812:
Looks good to me. Thanks!

Issue #898, opened 2015-01-15
tooltip should not disappear on data flow


J-F-Liu:
Tooltip should disappear only when mouse position is changed, otherwise the use don't have enough time to read the values if the flow speed is once per second.
If the data point on mouse position is chagned after flow then contents on tooltip can automatically update.

aendrew:
Confirming, just tried to read tooltips on the [flow example](http://c3js.org/samples/api_flow.html) and @J-F-Liu is right, they're impossible to read. Marking as bug.

masayuki0812:
This is intended, but it seems good to implement. So, please let me mark as enhancement.

J-F-Liu:
Find a solution now:
remove the line `$$.tooltip.style("display", "none");` in `c3_chart_internal_fn.redraw` function
remove the line `$$.hideTooltip();` in `c3_chart_internal_fn.generateFlow` function

aendrew:
@J-F-Liu Seems that would work — mind doing a PR? Thanks!

J-F-Liu:
RP created.

masayuki0812:
Merged your PR. I'll release this fix in the next version `0.4.11`. Thanks!!

aendrew:
This has been merged into master so I'm closing. Thanks to @J-F-Liu for the PR!

PR #1080, opened 2015-03-26
Keep tooltip on redraw and flow, only hide on mouse move out

J-F-Liu:
Related issue is [tooltip should not disappear on data flow #898](https://github.com/masayuki0812/c3/issues/898)


masayuki0812:
Merged. Thanks!!!

PR #1085, opened 2015-03-28
Set the X axis in subchart to visible or invisible 

vladsu:
I'm totally new in the github and all this stuff. Sorry if I do something wrong. 

I came across c3 for two days and found it suitable for me. Thanks for that!

But there are some points which I would like to improve. Here I have added two lines to the c3.js
and with following options the X axis could be set to hide
```
...
axis: {
  x: {
    show: false
  }
}
...
```
http://jsfiddle.net/vladsu/89szjf3r/3/ 

Best wishes,
Vlad

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -1077,6 +1077,7 @@
             subchart_show: false,
             subchart_size_height: 60,
             subchart_onbrush: function () {},
+            subchart_axis_x_show: true,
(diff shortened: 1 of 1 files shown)


masayuki0812:
Looks good to me. Thanks!

PR #1086, opened 2015-03-28
Fix data label position for zero value when all the data set is negative

yordis:
When all the data set is negative the label position of zero value will be above the the y axis line then you will not see the label.

[Click here for the JSFiddle](http://jsfiddle.net/yordisprieto/pse1wh23/)

@masayuki0812 I didn't do the Unit Test because I don't know where I should put the testing. I change two files and I need your help for know where I should put the Unit Testing for this.


yordis:
@masayuki0812 Sorry, is someone checking this?

masayuki0812:
@yordis Sorry for the late response. Merged and refactored a bit. Also I added a spec for this issue. Thanks!

yordis:
@masayuki0812 great! I didn't know about $$.hasPositiveValueInTargets :smile:

Issue #939, opened 2015-01-25
Performance Issues on create/resize


solgarius:
Hi,
I've noticed that c3.js can be quite slow when the window is resized or when the chart is initially generated. These issues are related to issue #730 which has been closed, however the performance is still quite poor.

In my app we have a number of charts on the same page (~7), and are planning on increasing this number. There are also several pages which means overall we have 50+ charts on our single page app (though mostly hidden). 

Initially (and preferably) I have the charts in variable sized containers and I'm finding a single resize will take the page 3-4sec. I've tried setting a fixed size.width/size.height and also the padding.left as in #730 however this only reduces the resize time to ~1.5sec (~200ms per chart). 

I have other charts in the same page that are using pure d3 and these all resize in around 10-20ms per chart. 

I've noticed in the profiler that the vast majority of the resize time (~180ms per chart) is spent in the function: getBoundingClientRect, which is called several times each resize and takes ~25ms per call.
At this stage I've only just started looking into what I can do to improve this performance, however I thought I'd raise the issue so if others have ideas for solutions or have seen the issue they can let me know. 

Overall I'm loving c3.js and its ease of use and features and really want to see how it progresses and am hoping I can find time to help contribute.

lilasquared:
Hey @solgarius, I posted something the other day about resize events and hiding / showing the charts.  Have you tried explicitly giving the chart options a height and width?  If the performance hit is during getBoundingClientRect it might be doing this to get the size.  If the size is set it might not need to do this call.  Just curious though.  Hope it helps!

solgarius:
@lilasquared, settting the height/width does seem to solve the problem, however I noticed that all of my 50+ charts are being resized even though only 7 are actively displayed on the screen. 

Ultimately, I'd prefer to allow c3 to figure out the resize for me and I don't have to explicitly set the size, hence the reason for this ticket. I.e. sure, I can work around it, but it would be better if we can resolve the core issue. 

Perhaps I'm looking at this issue a little too broadly. I suspect some of the problems should be resolved by #936, #926 and #857/#860 

I suggest though, to improve the performance further, is it possible to have new functions in the API: chart.active & chart.inactive (or similar) to explicitly notify the chart that it is currently not active and not to resize/do updates etc and then when set to active it can resize itself and do any other pending actions? 
This would differ from generate/destroy because it shouldn't require as large an overhead as these functions have.

lilasquared:
[...]

PR #1113, opened 2015-04-11
fixed performance issue while resizing and resize event removal on destroy

rknoll:
1. fixes the performance issue while resizing (based on #860)
  - further references: #857, #939, #926, #857
2. fixes the `destroy` method when using multiple charts on a page
  - `window.onresize` got set to `null` so all other charts did not resize anymore

### DIFF
--- a/src/api.chart.js
+++ b/src/api.chart.js
@@ -14,7 +14,22 @@ c3_chart_fn.destroy = function () {
     var $$ = this.internal;
 
     window.clearInterval($$.intervalForObserveInserted);
-    window.onresize = null;
+
+    if ($$.resizeTimeout !== undefined) {
+        window.clearTimeout($$.resizeTimeout);
+    }
+
+    if (window.detachEvent) {
+        window.detachEvent('onresize', $$.resizeFunction);
+    } else if (window.removeEventListener) {
+        window.removeEventListener('resize', $$.resizeFunction);
+    } else {
+        var wrapper = window.onresize;
+        // check if no one else removed our wrapper and remove our resizeFunction from it
+        if (wrapper && wrapper.add && wrapper.remove) {
+            wrapper.remove($$.resizeFunction);
+        }
+    }
 
     $$.selectChart.classed('c3', false).html("");
 
--- a/src/config.js
+++ b/src/config.js
@@ -7,6 +7,8 @@ c3_chart_internal_fn.getDefaultConfig = function () {
         padding_right: undefined,
         padding_top: undefined,
         padding_bottom: undefined,
+        resize_auto: true,
+        resize_timeout: 100,
         zoom_enabled: false,
         zoom_extent: undefined,
         zoom_privileged: false,
--- a/src/core.js
+++ b/src/core.js
@@ -319,20 +319,7 @@ c3_chart_internal_fn.initWithData = function (data) {
     }
 
     // Bind resize event
-    if (window.onresize == null) {
-        window.onresize = $$.generateResize();
-    }
-    if (window.onresize.add) {
-        window.onresize.add(function () {
-            config.onresize.call($$);
-        });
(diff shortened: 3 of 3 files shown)


aendrew:
Wait, this might fix #926? Somebody give this man a medal. :trophy: :+1: :shipit: 

Will evaluate and RTBC if it does as promised.

rknoll:
:+1:
hmm.. I just realized, there is still a problem if we have to use `window.onresize`.. the events won't get released correctly if you don't call `destroy` in the reversed order in which you created the graphs :sweat:

I'll try to find a solution for that (at least for the case where c3 is the only code that uses `window.onresize` on the page), stay tuned..

aendrew:
@rknoll Awesome, thanks for the update! 

Any idea whether it fixes how `generate()` keeps adding resize handlers when called on the same chart, leading to a massive memory hole over time? That's definitely something I want to test, but if it doesn't address that I won't bother. Thanks!

rknoll:
Definitely, as long as you call `destroy()` on the old one at least..
Otherwise you'd have to check if there is already a Chart initialized for that DOM element and call its `destroy()` function. Maybe by checking if it has the `c3` class on it?

jalebiman:
Hey all, thank you for this fix. Been struggling with this for about 2 weeks now and am very happy I stumbled upon this. This did indeed fix the leak when calling c3.generate repeatedly on the same chart. Thank you for the fix.

ergo:
Wow this is awesome, I need this :D

aendrew:
Here's the results of two benchmarks, first without calling `destroy()` between `generate()` calls and the latter with it:
no destroy calling

with destroy being called

[Plunker used](http://embed.plnkr.co/Xpu57IMAkDtxLiq9HXrN/preview)

Both show a steady rise in the number of listeners and ongoing memory leaks, though it seems like GC is working WAY better in the second instance. While it may not fix #926 entirely, it's definitely a step in the right direction.

---

That said, this isn't exactly what this PR is supposed to fix. It's supposed to be nullify the onresize events such that resizing the browser doesn't cause a memory leak (At least, if I'm understanding properly — @rknoll, please do let me know if I'm missing the point). The following graph shows me resizing the chart a bunch of times:

resizing

Maybe it's just because the buffer filled too quickly and all I've recorded is two resizes, but it seems like not a lot of spiking in the middle of the test, which is a good sign? Again, I'm really not the person who should be doing this, I really am not sure what I'm doing here... Will keep testing.

tl;dr — this PR seems to offer a pretty decent performance boost, and even if it doesn't resolve the aforementioned issues, it's definitely a step in the right direction and if nothing else gives us an internal API for continuing to attack resize-related issues. LGTM; :+1: for merge.

@rknoll That said, mind squashing your commits? Thanks!

aendrew:
I still vote to merge it, but I would be forever grateful if somebody with more of a clue what they're doing re: JavaScript performance profiling could take a look at it.

rknoll:
what I found is that if you manually trigger the GC in chrome, it will remove all listeners again.. I don't really have too much insights into the whole project to tell why it won't get freed by itself, and, I'm not a JS ninja either :-/
The main reason for this PR (at least for me) was that I have multiple C3 charts on a page, that get `destroy()`-ed and recreated individually, and the original implementation prevented me from destroying the graphs as it just removed the resize listeners completely (from all other graphs as well)..
Furthermore I really wanted to get #860 in as well to reduce the load on pages with more graphs or on mobile..
So even though it does not fix the "leaks" (without manually triggering GC), I'd also like to see this in the next release, at least for the 2 mentioned reasons :-) That said, I'll squash right now ;-)

masayuki0812:
I merged this PR. Thanks a lot! @rknoll 
One thing. I removed `resize.timeout` option because it's a kind of internal things and `100` seems enough for any case basically.

Issue #1054, opened 2015-03-09
Problem parsing date on parseDate function (when date is a string)


lbortoli:
I think I noticed the error:

c3_chart_internal_fn.parseDate = function (date) {
        var $$ = this, parsedDate;
        if (date instanceof Date) {
            parsedDate = date;
        } else if (typeof date === 'number' && !isNaN(date)) {// fixed wrong validation
            parsedDate = new Date(+date);
        } else {
            parsedDate = $$.dataTimeFormat($$.config.data_xFormat).parse(date);
        }
        if (!parsedDate || isNaN(+parsedDate)) {
            window.console.error("Failed to parse x '" + date + "' to Date object");
        }
        return parsedDate;
    };

The issue was on the validation
 if (typeof date === 'number' || !isNaN(date))

it should be
 if (typeof date === 'number' && !isNaN(date))

michaelrodov:
fix commited

PR #1121, opened 2015-04-16
Fix for issue #1054

lbortoli:
This is the fix for the issue "Problem parsing date on parseDate function when date is a string".
The validation when 'date' is a number was wrong.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -983,7 +983,7 @@ c3_chart_internal_fn.parseDate = function (date) {
         parsedDate = date;
     } else if (typeof date === 'string') {
         parsedDate = $$.dataTimeFormat($$.config.data_xFormat).parse(date);
-    } else if (typeof date === 'number' || !isNaN(date)) {
+    } else if (typeof date === 'number' && !isNaN(date)) {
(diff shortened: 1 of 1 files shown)


aendrew:
LGTM! :+1: That said, is it necessary to check that it's not NaN when it just checked that the same variable is a number?

lbortoli:
I think you are right,
I guessed it wouldn't hurt to have that validation also :P, but I leave that decision for you guys...
Thanks for the reply!

masayuki0812:
@lbortoli Thanks for the PR, but I'm not sure the situation. Can you provide an example where we get an error?

lbortoli:
Assume 'date' has value NaN, if the condition remains as:
if (typeof date === 'number' || !isNaN(date)) {
   parsedDate = new Date(+date);
}
the condition would be true since NaN is of type 'number', and new Date(+date) fails. So we need to validate !isNaN(date) as well, that's why I changed '||' for '&&'.
Does it make sense?

masayuki0812:
@lbortoli It's true. You're right. Thanks!!

PR #1138, opened 2015-05-04
Disabling the gray color highlighting on click in iOS devices

rsudhakar:
When any graph is opened in iOS devices, clicking on it highlights the whole graph with gray color briefly. This hinders the interactivity as it does it for the whole graph area. 

The commit adds a css style that will disable this. 

See this for more information:
https://developer.apple.com/library/mac/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html#//apple_ref/doc/uid/TP40006510-SW5


rsudhakar:
Creating another pull request with a specific branch for this change

Issue #1141, opened 2015-05-04
Ability to increase the clickable area of points in the additional y-axis chart


rsudhakar:
When an additional y-axis chart is opened in iPad, it is not responsive enough for the data point click handler to trigger when tapped on it. 

This is because the minimum distance is hard coded and cannot be configured. Can this be exposed to be configurable when calling c3.generate method?

Code reference:

data.js
```
c3_chart_internal_fn.findClosest = function (values, pos) {
       var $$ = this, minDist = 100, closest;
```

interaction.js
```
        // Show cursor as pointer if point is close to mouse position
        if ($$.isBarType(closest.id) || $$.dist(closest, mouse) < 100) {

        // select if selection enabled
        if ($$.isBarType(closest.id) || $$.dist(closest, mouse) < 100) {
```

PR #1142, opened 2015-05-04
Patch for #1141

rsudhakar:
Have modified the data.js and interaction.js files to calculate the distance between the two points and compare it with the configured value (which is passed to c3.generate method along with other config)

 Example

         point: {
             show: true,
             r: 5,
             sensitivity: 20
         },

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -1158,6 +1158,7 @@
             // point - point of each data
             point_show: true,
             point_r: 2.5,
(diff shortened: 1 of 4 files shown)


aendrew:
LGTM, though haven't done a manual review. :+1:

masayuki0812:
LGTM, too. Thanks!!

PR #1147, opened 2015-05-06
Add chart.api ($$.api) to 'onunselected' callback

lbortoli:
It should behave as the 'onselected' callback.


masayuki0812:
You're right. Thanks!

lbortoli:
Actually, I think it would be better to pass 'this' as a parameter for ALL the callbacks, instead of passing 'this.api', since we may have to use other components such as 'color' function for example...
Would you add that change in all the callbacks C3 provides? What do you think?

PR #1153, opened 2015-05-07
Enforcing the AMD factory standard

admwx7:
https://github.com/amdjs/amdjs-api/blob/master/AMD.md designates that the factory should be a function.


masayuki0812:
You're right. Thanks!

PR #1177, opened 2015-05-22
updating d3

### LINKED ISSUE #962: Error installing v0.4.9 via NPM
I do not use bower for browser package management but use npm instead. When I do `npm install c3@0.4.9` or `npm install c3@0.4.8` they both will give installation error which I detailed inside brianmcd/contextify#153.

luisrudge:
fixes #962

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "gitHead": "84e03109d9a590f9c8ef687c03d751f666080c6f",
   "readmeFilename": "README.md",
   "dependencies": {
-    "d3": "<=3.5.0"
+    "d3": "^3.5.5"
   },
   "devDependencies": {
     "grunt": "^0.4.5",


luisrudge:
just by updating the version, the install runs like a dream:

```
c:\git\rr\RealRoi.Cis.Frontend\src (master)
λ npm i -S luisrudge/c3
npm WARN package.json package@0.0.0 No description
npm WARN package.json package@0.0.0 No repository field.
npm WARN package.json package@0.0.0 No README data
c3@0.4.10 node_modules\c3
└── d3@3.5.5
```

aendrew:
We've had some issues in the past with updating D3; while I'm all in favour of updating D3 and getting rid of the JSDOM dependency, it's probably necessary to test that particular version of D3 before merging this.

@masayuki0812 We should really create a test suite specifically for D3 versioning functionality, containing various regressive bugs we've noticed when updating D3. Maybe even use something like E2E testing with Selenium WebDriver?

luisrudge:
Yeah.. I have a fork with an updated version of d3 and tooltips aren't working. :(

masayuki0812:
jasmine cannot be used? I think it's better to separate test if possible. I'll try later anyway.

luisrudge:
@masayuki0812 have you tried that?

luisrudge:
any thoughts on this? I can't install c3 on windows with the old dependency

aendrew:
@masayuki0812 Jasmine's generally used for unit testing, less-so for E2E testing. You're right though, we could probably just use Jasmine for this and add a bunch more test runners in Travis (I.e., have tests also automatically run for Firefox, Chrome, Safari and IE, instead of just PhantomJS, which would provide a bit more certainty that new version of D3 aren't breaking things). 

How we'd go about writing a test suite that encapsulates all ongoing changes to D3 is another question, would probably necessitate following the changelog pretty closely. Open to any suggestions!

aendrew:
As per the discussion starting [here](https://github.com/masayuki0812/c3/pull/1166#issuecomment-131826120), I'm increasingly unconvinced that constraining D3 to a specific bugfix version is the best idea.

aendrew:
Closing because this is constrained to the minor release version, and we're having enough issues with just constraining D3 to the bugfix version. Let's discuss this at #1166.

PR #1188, opened 2015-05-31
Add the ability to prevent elements from the showing up in legend and tool-tip

j-white:
As the title suggests, we're trying to prevent specific series from showing up in the legend and too-tips.

This patch allows series to be hidden from these when their names are set null via the data_names option.

I'll be happy to update the patch if there's an alternative approach that's preferred.


masayuki0812:
LGTM. Thanks!

Issue #1189, opened 2015-06-01
Clip path rect receiving no height/width (Chart appears blank)


impatient:
We're using c3 in an angular app, however when we draw the sample donut chart, it simply shows white space.  The root causes was that the defs + rects don't have a proper width/height associated with them.

![image](https://cloud.githubusercontent.com/assets/3408099/7914550/4c18e534-083d-11e5-8b7f-e2bb1e98f69f.png)

This is an odd situation. This works in most environments and the test case that I've created actually passes with the current code.

When determining size, it looks like the d3 select all doesn't do DOM traversal, so I wouldn't expect it to work with just id's as an array as written.

jeremy-walsh-vgh:
seeing the same issue from a chart embedded on a visualforce page in salesforce.  rect has no height/width attributes.  Strange thing for me is that the page I'm debugging was working for a some time and then stopped; don't know what changed to stop it from rendering, but if I delete the clip path node from the defs in the chrome inspector, the chart renders correctly.

NemSavic:
@jeremy-walsh-vgh if you're working with Visualforce pages in Salesforce it looks like this became an issue as of the Salesforce Winter 16 release. When you use apex:form tags or have showHeader="true" set in the apex:pages tag, Salesforce will load their own JS files which are now likely conflicting with something in C3.js

impatient:
@jeremy-walsh-vgh I'd forgotten about this.  We ended up switching to D3 and the issue only recently popped up again.  I bet the updated Visualforce pages include Sizzle.. globally, which causes D3  to go down this path.

https://github.com/mbostock/d3/blob/master/src/selection/selection.js#L24

I bet if you remove Sizzle from the window while bootstrapping your code you'd be good to go.

Sample Sizzle bug that could possible be introduced.

https://github.com/jquery/sizzle/issues/322

Please update if it works, I'm closing this, since it's an issue with D3/Sizzle.  Good luck.

PR #1190, opened 2015-06-01
Switch selectAll syntax from array syntax to comma separated.  #1189

impatient:
Our code matches the donut sample pretty much verbatim, but we get nothing displayed without the change in the pull request. 

http://c3js.org/samples/chart_donut.html

Test code that needs a little bit of cleanup.  It feels like it's testing a side effect.  Can clean it up if you've got any suggestions.

```
        it('Clip paths should have a width', function() {

            var defs = d3.select('#chart svg').select('defs');

            //can't select clip path via d3 select (at least in phantom)
            var clips = [].reduce.call(defs.node().childNodes, function(memo, child) {
                expect(d3.select(child).select('rect').attr('width')).toBeDefined();
                memo.push(child);
                return memo;
            }, []);

            clips = d3.selectAll(clips);

            clips.selectAll('rect').attr('width', null);

            clips.each(function() {
                expect(d3.select(this).select('rect').attr('width')).toBeNull();
            });

            chart.internal.updateSvgSize();

            clips.each(function() {
                expect(d3.select(this).select('rect').attr('width')).toBeDefined()
            });

            expect(clips.size()).toBe(5);
        });

```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -856,7 +856,7 @@ c3_chart_internal_fn.updateSvgSize = function () {
     var $$ = this,
         brush = $$.svg.select(".c3-brush .background");
     $$.svg.attr('width', $$.currentWidth).attr('height', $$.currentHeight);
-    $$.svg.selectAll(['#' + $$.clipId, '#' + $$.clipIdForGrid]).select('rect')
+    $$.svg.selectAll(['#' + $$.clipIdForGrid,'#' + $$.clipId].join(',')).select('rect')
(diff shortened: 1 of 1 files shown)


masayuki0812:
I cannot see the problem this PR fixes. Can you tell us more detail such as browser and code version?

masayuki0812:
Referred issue has been closed, so let me close this PR as well. Thank you.

PR #1193, opened 2015-06-03
Fixes #1192

### LINKED ISSUE #1192: Add onshow and onhide events for tooltip
Should be easy enough to provide show and hide events for tooltip

GermainBergeron:
Added onshow and onhide functions on tooltip


masayuki0812:
LGTM. Thanks!

Issue #687, opened 2014-11-07
Not compatible with Touch


sjunnesson:
It seems that c3js is using " mouse = d3.mouse(this);" everywhere in the code to handle events. This makes touches from mobile devices undetectable. Could a  "touch = d3.touch(this); " be implemented in some way to handle this?

nillotrejos:
data.onclick is working on Android devices but is not responding on iOS devices.
When clicking on iOS, the tooltip displays, but onclick callback is never triggered.

nillotrejos:
I kept looking at the code and playing with the graphs in iOS simulator and noticed that if you set the tooltip to false, the data.onclick callback works as expected on iOS.

http://codepen.io/nillotrejos/pen/jEXNLR

akoinesjr:
I had the same issue on an iPhone 4S and was able to "fix" it by removing the tooltip as suggested by @nillotrejos.

jamesgarfield:
I'm having a similar experience with iOS TouchEvent clicks not working.

By adding some logs to the event handlers I can see that on desktop and Android a Touch produces the following events in this order from the "generateEventRectsForSingleX" series of events

* MouseOver
* MouseMove (triggers tooltip)
* Click

However, on iOS devices, the only events that are produced are

* MouseOver
* MouseMove (triggers tooltip)

I have *not* unilaterally had the same success with fixing this by disabling tooltips (nor is that really an option in our case)

mcarey1590:
has anyone found a fix for this. On safari(iOS) when i tap on a data point it shows the tooltip, but it does not fire the onclick event. On chrome(Android) when i tap on a data point it works as expected and fires the onclick event, but there is no way to view the tooltip unless you have a device that supports hover events

jamesgarfield:
@mcarey1590 I have a PR out about this that worked for us. It was a long time ago, so I don't know how it aligns with the current code base, but it should at least provide a starting point for you.

https://github.com/masayuki0812/c3/pull/1206

mcarey1590:
Thanks I will look into it

eljog:
Hi,

I'm using C3 js in my Cordova Hybrid app. I have multple charts on the screen, so the user is supposed to scroll.
The chart works well in android but the scrolling is not smooth in iphone. Tested on iPhone 6.
Kindly help me if there is any tip make the scrolling smooth in iPhone.

Thanks,
Eljo

Suffren:
[...]

PR #1206, opened 2015-06-10
Add 'tap' TouchEvent to fix iOS selection bug. #687

jamesgarfield:
Currently 'click' events do not always work on iOS devices. 

This is due to a combination of a subtlety in the ordering and conditions in which events are fired, and what chart effects each event can trigger in a chart.

When a `mousemove` event causes the page content to change, the simulated `click` event that you normally get doesn't fire on iOS. See [Apple Documentation](https://developer.apple.com/library/mac/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html) between Fig. 6-3 and Fig 6-4 for more details.

This means that if you chart uses tooltips, or the guideline, these actions can cause the `click` event to get canceled by virtue of having changed the page content.

This change extracts the `click` functionality in both single and multi X event rectangles to common functions that can be used by both the `click` event and a simple, standard implementation of a `tap` event that is generated by monitoring `touchstart` and `touchend` for tap-like behavior.

Changes tested on 
iPhone 4 
iPhone 6
iPad 2
Android 4.0
Android 4.4
Chrome (Desktop & Android)
Firefox (Desktop)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -214,6 +214,9 @@ c3_chart_internal_fn.getDefaultConfig = function () {
             left: 0
         },
         title_position: 'top-center',
+        //TouchEvent configuration
+        touch_tap_radius : 20,  //touch movement must be less than this to be a 'tap'
+        touch_tap_delay : 500,  //clicks are suppressed for this many ms after a tap
     };
 
     Object.keys(this.additionalConfig).forEach(function (key) {
--- a/src/interaction.js
+++ b/src/interaction.js
@@ -109,7 +109,27 @@ c3_chart_internal_fn.updateEventRect = function (eventRectUpdate) {
         .attr("height", h);
 };
 c3_chart_internal_fn.generateEventRectsForSingleX = function (eventRectEnter) {
-    var $$ = this, d3 = $$.d3, config = $$.config;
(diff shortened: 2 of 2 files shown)


OldrichKruchna:
I'm facing the same problem, so I've tried to build your PR and to include it to my project, but it didn't help. I've tested it on Android 4.4 and Ipad 2. Weird thing is that example on c3.js webpage actually works for me on both devices.

aendrew:
This looks useful. LGTM, :+1:.

masayuki0812:
Sorry, I could not release this fix in 0.4.11. This should be included in the next version.

aendrew:
Merged into `dev`: 353c770

lmmfranco:
Is 0.4.12 coming anytime soon?

maZahaca:
Hello
It's a bit time left since "Jun 10, 2015" can we merge this and use this chart for mobile?

ArmindoMaurits:
This change is still hanging loose somewhere. Can this get merged into dev ASAP? Not being able to click on a column while having tooltips on IOS is pathetic...

cdomergue:
Still nothing ?

PR #1207, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


aendrew:
LGTM :+1: I did this with Axis, it's so great not having another place to forget to bump version numbers....

masayuki0812:
Thanks!

Issue #1208, opened 2015-06-10
Legend event method customization does not provide consistent behavior


jamesgarfield:
When customizing the methods for a legend, providing a new `onclick` method completely overrides the default click behavior
e.g.
```
 if (config.legend_item_onclick) {
            config.legend_item_onclick.call($$, id);
} else {
            //...
}
```

However, the corresponding `mouseover` and `mouseout` events do not conform to this same behavior, they instead perform their default action AND the custom action.
e.g.
```
$$.d3.select(this).classed(CLASS.legendItemFocused, true);
if (!$$.transiting && $$.isTargetToShow(id)) {
            $$.api.focus(id);
}
if (config.legend_item_onmouseover) {
            config.legend_item_onmouseover.call($$, id);
}
```

This makes it difficult to truly provide a custom implementation of the legend since you are always getting some of the default behavior. Specifically this makes customizing the behavior on a touch-screen device difficult because you do not get the same `mouseover`->`mouseout` transition you get with mouse-based devices.

PR #1209, opened 2015-06-10
Change legend method customization to be consistent across all methods

jamesgarfield:
Ref #1208 

Alter how custom methods work in the `mouseover`/`mouseout` legend events to be consistent with how `click` works.

### DIFF
--- a/src/legend.js
+++ b/src/legend.js
@@ -240,19 +240,23 @@ c3_chart_internal_fn.updateLegend = function (targetIds, options, transitions) {
             }
         })
         .on('mouseover', function (id) {
-            $$.d3.select(this).classed(CLASS.legendItemFocused, true);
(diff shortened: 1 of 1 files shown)


masayuki0812:
You're right. Thank you!!!

DemarsM:
Is it normal that legend item onmouseover function overide the default function wich focus on the graph item element?

Since I have defined a custom onmouseover function, the focus is not triggered anymore.

jamesgarfield:
@DemarsM It's been a while sine I looked at this, but yes, I believe that is the desired functionality. The entire point is to be able to say "I don't want the default functionality, I want my own". The apis to force focus exist, so if you want custom behavior & default highlight, you would beed to add the highlighting code to your custom handler.

DemarsM:
Ok thank you!

Issue #131, opened 2014-04-07
Firefox fails to draw charts in jquery UI tabs


lblb:
Hello,
I noticed a strange issue with firefox when using your chart library inside tabs of jquery UI. The chart is rendered correctly on the first tab, however it fails completely on other tabs. I'm using tabsactivate event to create a chart, at this moment tab panel is fully visible. While investigating I see it throws exception NS_ERROR_FAILURE excecuting function getBBox()  inside getMaxTickWidth() at line 784: 

```
    function getMaxTickWidth(id) {
        var maxWidth = 0, axisClass = id === 'x' ? CLASS.axisX : id === 'y' ? CLASS.axisY : CLASS.axisY2;
        d3.selectAll('.' + axisClass + ' .tick text').each(function () {
            var box = this.getBBox();
            if (maxWidth < box.width) { maxWidth = box.width; }
        });
        return maxWidth < 20 ? 20 : maxWidth;
    }
```

I'm using latest firefox v28, however IE9-11 and latest chrome work fine in this case. Please take a look at live example here: http://jsfiddle.net/Lm2e5/4/
Line 784 is not the only one where getBBox() is used, so if I configure to hide axis, exception will be rised in a different place of the code.

I tracked down that jquery UI is not responsible for this exception, because it can be reproduced by toggling visibility of adjanced DIVs. The key is to create a chart and hide it with visibility:none. Then set visibilty:block to nearby DIV and draw a new chart, while the first one is still there hidden in the DOM.

As a temporary workaround I replaced getBBox()  with getBoundingClientRect(). It seems to work fine in Firefox, but I can't understand why getBBox() fails. For instance another D3 based charting library called Dimple also uses getBBox function, but it's charts work well inside tabs.

PR #1226, opened 2015-06-19
Fix firefox NS_ERROR_FAILURE issue caused by the new titles feature

laurence-hudson-tessella:
NS_ERROR_FAILURE when rendering titles in off-screen chart in firefox only. This is a reintroduction of same issue in #131. Applied same fix as cd004fea.


laurence-hudson-tessella:
Superseded by #1231.

PR #1230, opened 2015-06-23
Fix #1229 - Incorrect x tick positioning

### LINKED ISSUE #1229: Incorrect x tick positioning, for some data ranges
See http://codepen.io/anon/pen/xGPVgJ?editors=001
Correct tick values, but placed at wrong position on x axis.

laurence-hudson-tessella:
Fix #1229 - Incorrect x tick positioning

``Axis.prototype.generateTickValues`` incorrectly assumes that the input ``values`` (unique x values) will be numerically/date sorted. These values are extracted by ``c3_chart_internal_fn.mapTargetsToUniqueXs`` which uses ``d3.set.values``. The order of the output of ``d3.set.values`` is [not guaranteed](https://github.com/mbostock/d3/wiki/Arrays#set_values) (it happens to be string alpha sorted). In many cases this does not matter, but if the min or max numeric value get put in the wrong position the erroneous behavior is seen.

I have simply added a guard in ``c3_chart_internal_fn.mapTargetsToUniqueXs``, to numerically sort the unique x values before passing into ``Axis.prototype.generateTickValues``.

Before: http://codepen.io/anon/pen/xGPVgJ?editors=001
After: http://codepen.io/anon/pen/jPaqGB?editors=001

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -171,7 +171,8 @@ c3_chart_internal_fn.filterTargetsToShow = function (targets) {
 c3_chart_internal_fn.mapTargetsToUniqueXs = function (targets) {
     var $$ = this;
     var xs = $$.d3.set($$.d3.merge(targets.map(function (t) { return t.values.map(function (v) { return +v.x; }); }))).values();
-    return $$.isTimeSeries() ? xs.map(function (x) { return new Date(+x); }) : xs.map(function (x) { return +x; });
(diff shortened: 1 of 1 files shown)


aendrew:
@laurence-hudson-tessella LGTM! :+1: :shipit: Marking as RTBC/Ready.

masayuki0812:
LGTM. Thanks!!

PR #1231, opened 2015-06-23
Fix #1225 - (0.4.11-rc1) Incorrect layout when using titles on charts that are not visible at generation

### LINKED ISSUE #1225: (0.4.11-rc1) Incorrect layout when using titles on charts that are not visible at generation
Example here: http://codepen.io/anon/pen/aOLwGd.
Title is above the top of the SVG.

0.4.11-rc1 - 4b34a19 
#1199

Note: codepen example doesnt work at all in firefox due to #1160(#131). but title bug can be seen in chrome. Firefox specific issue is resolved by PR #1226 (http://codepen.io/anon/pen/aOLyEL), but layout issue still present.

laurence-hudson-tessella:
* Use getTextRect to measure title size, 
* Fix getTextRect so that any custom css font used is applied while measuring (also applies to legend)

Also fixes the firefox "NS_ERROR_FAILURE" regression, as getBBox is no longer used. As such this supersedes #1226.

A slightly cleaner implementation is at [laurence-hudson-tessella/c3/fix-title-layout-clean](https://github.com/laurence-hudson-tessella/c3/tree/fix-title-layout-bug-clean), but it breaks some specs (only when run in node?) for reasons that are not immediately clear.

Before: http://codepen.io/anon/pen/aOLwGd (wont work in firefox)
After: http://codepen.io/anon/pen/VLrjev

### DIFF
--- a/src/legend.js
+++ b/src/legend.js
@@ -130,7 +130,7 @@ c3_chart_internal_fn.updateLegend = function (targetIds, options, transitions) {
 
     function getTextBox(textElement, id) {
         if (!$$.legendItemTextBox[id]) {
-            $$.legendItemTextBox[id] = $$.getTextRect(textElement.textContent, CLASS.legendItem);
+            $$.legendItemTextBox[id] = $$.getTextRect(textElement.textContent, CLASS.legendItem, textElement);
         }
         return $$.legendItemTextBox[id];
     }
--- a/src/text.js
+++ b/src/text.js
@@ -47,14 +47,16 @@ c3_chart_internal_fn.redrawText = function (xForText, yForText, forFlow, withTra
             .style("fill-opacity", forFlow ? 0 : this.opacityForText.bind(this))
     ];
 };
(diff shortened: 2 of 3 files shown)


aendrew:
Yus! LGTM :+1:. Re: the breaking tests, seems to be passing at the moment. Restarting the build in Travis, will see if it passes again...

laurence-hudson-tessella:
I think you miss understand, the PR works fine. 
However it should be possible to implement it in a cleaner way, [which I did](https://github.com/laurence-hudson-tessella/c3/tree/fix-title-layout-bug-clean), but it [fails some specs](https://travis-ci.org/laurence-hudson-tessella/c3/builds/69325150) for reasons that are not clear.

aendrew:
@laurence-hudson-tessella Ahhh, okay, makes sense. Will take a look at the alternative implementation tomorrow, see if I can't figure out what's breaking.

masayuki0812:
Looks very nice. Thank you!!

Issue #1246, opened 2015-07-01
Extra x-axis appearing upon resize


Palladian1881:
Hi,

Here is my graph. It is required to be resizable:
![image](https://cloud.githubusercontent.com/assets/8864735/8462907/10ffc5fc-1ff4-11e5-9a70-2e9df9d4cce0.png)

When I resize it, however, another x-axis appears:
![image](https://cloud.githubusercontent.com/assets/8864735/8462917/28b51d64-1ff4-11e5-9c89-28942c2277b0.png)

Here is my code. Thanks!

    <head>
    <script>
    var randomBetween = function(min,max){
         return Math.random() * (max-min) + min;
    }
    $(function() {
        $( "#windowDiv" ).resizable();
        var chart = c3.generate({
             bindto: '#chart',
              data: {
                 columns: [
                          ['Reports_Generated', randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500),randomBetween(0,500)],
                          ],
                 type: 'spline',
             }
    });

    $("#windowDiv").resize(function(){
             chart.resize({
                 height: $("#windowDiv").height() - 40,
                 width: $("#windowDiv").width() - 20
             });
        });
    });
  
    </script>
    </head>
    <body>
    <div id="windowDiv" style="width:404px;height:414px;border:1px  solid black;">
    <div id="chart"></div> 
    </div> 
    </body>

Palladian1881:
Turns out that the second x axis was always there, it was just hidden initially. I fixed the problem by removing the duplicate axis:
    $('.c3-axis-x').eq(1).remove();

Not a great solution, but it works.

aendrew:
I'm experiencing this too after updating to 0.4.11-rc1 in Axis, and I don't even need to resize the browser window. Investigating a bit further, it's actually the subchart x-axis, and changing the style property from `visibility: hidden` to `display: none` hides it. I'm not sure what the issue is; if [I just rip that bit of code out and put it in a jsFiddle](http://jsfiddle.net/aendrew/nypacLt9/), it's still visible, even without any CSS to override it or anything. Rather confused, behaviour persists in both Firefox and Chrome.

aendrew:
[...]

PR #1251, opened 2015-07-03
Makes subchart x-axis visibility dependent upon subchart visibility

aendrew:
Currently it's [possible to see the subchart x-axis](http://jsfiddle.net/aendrew/Lr6dbgwb/5/) if `subchart.visible` is `false` but `subchart.axis.x.visible` is `true` (and the SVG area gets extended a bit). See #1246.

This PR short-circuits the conditional for `subchart.axis.x.visible` if `subchart.visible` is falsy.

### DIFF
--- a/src/subchart.js
+++ b/src/subchart.js
@@ -40,7 +40,7 @@ c3_chart_internal_fn.initSubchart = function () {
         .attr("class", CLASS.axisX)
         .attr("transform", $$.getTranslate('subx'))
         .attr("clip-path", config.axis_rotated ? "" : $$.clipPathForXAxis)
-        .style("visibility", config.subchart_axis_x_show ? 'visible' : 'hidden');
+        .style("visibility", config.subchart_show && config.subchart_axis_x_show ? 'visible' : 'hidden');
 };
 c3_chart_internal_fn.updateTargetsForSubchart = function (targets) {
     var $$ = this, context = $$.context, config = $$.config,


daxfohl:
I was about to make the same pull request when I saw this one.  Was thinking a more flexible option though would be to use `context.style('visibility')` instead of `config.subchart_show`, just in case there's some higher-level container in the future.  Currently my fix looks like

```
c3_chart_internal_fn.initSubchart = function () {
    var $$ = this, config = $$.config,
        context = $$.context = $$.svg.append("g").attr("transform", $$.getTranslate('context')),
        visibility = config.subchart_show ? 'visible' : 'hidden';

    context.style('visibility', visibility);

    // Define g for chart area
    context.append('g')
        .attr("clip-path", $$.clipPathForSubchart)
        .attr('class', CLASS.chart);

    // Define g for bar chart area
    context.select('.' + CLASS.chart).append("g")
        .attr("class", CLASS.chartBars);

    // Define g for line chart area
    context.select('.' + CLASS.chart).append("g")
        .attr("class", CLASS.chartLines);

    // Add extent rect for Brush
    context.append("g")
        .attr("clip-path", $$.clipPath)
        .attr("class", CLASS.brush)
        .call($$.brush);

    // ATTENTION: This must be called AFTER chart added
    // Add Axis
    $$.axes.subx = context.append("g")
        .attr("class", CLASS.axisX)
        .attr("transform", $$.getTranslate('subx'))
        .attr("clip-path", config.axis_rotated ? "" : $$.clipPathForXAxis)
        .style("visibility", config.subchart_axis_x_show ? visibility : 'hidden');
};
```

aendrew:
Closing because #1294 is a better implementation.

PR #1253, opened 2015-07-03
Adds title.author, title.source, title.spacing. Resolves #1252.

### LINKED ISSUE #1252: Allow titles to have formatting
I'm not sure if I'm either over- or under-thinking this, but it'd be cool if the new title feature could be formatted a bit (I'm in a situation where I'd want titles in the format "**Chart Name** — Ændrew Rininsland/The Times"). You could do something like supply `<tspan class="chart-name">Chart Name</tspan> — Ændrew Rininsland/The Times` as the value of `config.title.text`, but that will print as plaintext due to [ln. 4 of title.js](https://github.com/masayuki0812/c3/blob/master/src/title.js#L4).

I thought it'd be as easy as changing that call from `d3.selection.text()` to `d3.selection.html()`, but even though it works in practice, 5/7 of the tests in title-spec.js fail (Oddly, the `d3.select('.c3-title').node().textContent` comes out as empty string.), and I really have no idea why.

@davidpoore Any ideas?

aendrew:
Adds author and source to titles, wraps title.text in a tspan with class `.c3-title-text`. Author and source are styleable via `.c3-title-author` and `.c3-title-source`. Renders SVG as:

```svg
<text class="c3-title">
  <tspan class="c3-title-text">{{title.text}}</tspan>
  <tspan class="c3-title-author" dx="{{config.title.spacing}}">{{title.author}}</tspan>
  <tspan class="c3-title-source" dx="{{config.title.spacing}}">{{title.source}}</tspan>
</text>
```

Config changes:

+ config.title.author (default: `undefined`)
    +  The chart's ostensible author.
+ config.title.source (default: `undefined`)
    + The ostensible source of the chart's data
+ config.title.spacing (default: `5`)
    + Distance between the elements.


masayuki0812:
`author` and `source` are too specific for some situation and it should be a plugin probably.

aendrew:
@masayuki0812 That's a good point. Will release this as a plugin to do that, if nothing else would be a good opportunity to figure out how to make the title system pluggable. :+1:

PR #1273, opened 2015-07-14
Fixed bug with tooltip names not being processed with $$.getTargetSel…

alycklama:
…ectorSuffix, resulting in not being able to style tooltips correctly


masayuki0812:
You're right. Thanks!!!

Issue #1065, opened 2015-03-16
Error in getShapeOffset()


craigsoules:
This bug is seen in RC5 for 0.4.10

If the data.xs is used, and the two series being plotted have different numbers of x-values provided (e.g., there are 5 data points for series A and 4 for series B) then there is an error seen in the getShapeOffset() function.  Below is a fix that not only confirms that there are enough points, but that the values align.  Definitely not the most efficient solution, but hopefully will lead you to a better fix.

```
-                    if (values[i].value * d.value >= 0) {
-                        offset += scale(values[i].value) - y0;
+                    // check if the x values line up
+                    if (typeof values[i] === 'undefined' ||
+                        (values[i].x !== d.x) && (values[i].x - d.x !== 0)) {
+                        // if not, try to find the value that does line up
+                        i = -1;
+                        for (var j in values) {
+                            if ((values[j].x === d.x) || (values[j].x - d.x === 0)) {
+                                i = j;
+                                break;
+                            }
+                        }
+                    }
+
+                    if (i in values) {
+                        if (values[i].value * d.value >= 0) {
+                            offset += scale(values[i].value) - y0;
+                        }
```

bryyyon:
Thank you for this, @craigsoules!  This fixes the alignment problem that can also happen with passing equal-length, grouped series data points to flow() in stacked bar charts. Issue #1068.  Posting here too in effort to help anyone else with that issue.

ggamir:
Here's a jsfiddle exemplifying the issue http://jsfiddle.net/9rm1qqmm/2/

Uncomment the four commented lines for the error.

masayuki0812:
I merged https://github.com/masayuki0812/c3/pull/1275 and this will be released in 0.4.11 shortly!

PR #1275, opened 2015-07-16
Allow data series with unequal amount of x-values

ggamir:
This is in reference to issue #1065 and proposing a pull request for the fix by @craigsoules . Currently, if there are data series that have an unequal amount of x-values, then there is a possibility that the code will error by trying to access an index out of bounds. This fix allows uneven data sets.


masayuki0812:
LGTM. Thank you very much!!! @ggamir @craigsoules

PR #1281, opened 2015-07-22
XHR headers with data.url

shukbot:
Allows for an optional data.headers object to be sent with the request


aendrew:
Good idea — could you please do the following?

a. Add a unit test. Given it's header, it's a bit more difficult, but you should be able to do so with Jasmine mocks. 

b. Please don't bother concatenating and minifying the dist version; we generally only do that during releases. Granted, I'm questioning the wisdom of this suggestion; see [this Gitter post](https://gitter.im/masayuki0812/c3?at=55b6168c61d39fd716068e25).

c. As always, please squash your commits.

Many thanks!

masayuki0812:
@aendrew Thanks for the comment, but merged for the moment. Let's add spec later.

PR #1285, opened 2015-07-23
Fixing firefox issue (mozilla bugzilla: #612118) for title rendering.

upphiminn:
Firefox has an issue with getting the bounding box for non-attached nodes or attached but hidden nodes. 

See https://bugzilla.mozilla.org/show_bug.cgi?id=612118 for more info. Right now this throws an exception and breaks the chart generation.

I am re-throwing other exceptions except this for now but I think I might change it. The whole chart generation shouldn't break because of the title :)

@masayuki0812

### DIFF
--- a/src/title.js
+++ b/src/title.js
@@ -11,19 +11,39 @@ c3_chart_internal_fn.redrawTitle = function () {
           .attr("y", $$.yForTitle.bind($$));
 };
 c3_chart_internal_fn.xForTitle = function () {
-    var $$ = this, config = $$.config, position = config.title_position || 'left', x;
+    var $$ = this, config = $$.config, position = config.title_position || 'left', x, textWidth;
+    try {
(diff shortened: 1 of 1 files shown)


masayuki0812:
Thank you for the PR, but this could have been fixed by this PR https://github.com/masayuki0812/c3/pull/1231 . Can you check again with `0.4.11-rc2` ? Thanks.

upphiminn:
Interesting, haven't seen that PR. It does look to solve the same issue. I think you should merge that one, I'll close this. 

I might have a different PR soon with another issue in FF (something to do with charts in hidden iframes :smiley:  )

masayuki0812:
Thanks. I already merged #1231 , so this can be closed now. And another PR is also welcome :smiley:

aendrew:
Closing as duplicate.

PR #1289, opened 2015-07-24
Added classname as a config parameter which will set svg class

dpiccone:
I had to introduce this for a project I am working on, maybe it can be added to the API


aendrew:
:+1: Not a bad idea at all. Could use a unit test though — mind adding one? Also, when squashing your commits, please remove c3.js because that will cause merge conflicts (It generally gets built during new releases. Thanks!

dpiccone:
I added tests and removed the minified c3 :)

Gruntfile.coffee
```diff
@@ -89,7 +89,8 @@ module.exports = (grunt) ->
           c3:
             src: 'c3.js'
             options:
-              specs: 'spec/*-spec.js'
+                #specs: 'spec/*-spec.js'
+              specs: 'spec/core-spec.js'
```
masayuki0812: Can you fix this?

masayuki0812:
@dpiccone Thank you for the PR. LGTM basically, but can you change the config to `svg_classname`? It seems easier to understand which classname it is :)

dpiccone:
Cleaned up Gruntfile and changed option name

dpiccone:
Updated the config, fixed tests

aendrew:
@dpiccone I'm working through all the PRs and figure this is a good candidate, but would you mind please squashing your commits? Thanks! RTBC otherwise! :+1:

masayuki0812:
Thank you!!

PR #1294, opened 2015-07-28
Bugfix 1292, 1246 by making subchart axis hidden if subchart is hidden

daxfohl:
This fixes the issue created in masayuki0812/c3@1168b69#diff-a88a43467d0fb57fc0542b3e9341cdf2 where the subchart axis is shown even if the subchart is hidden.

Fixes https://github.com/masayuki0812/c3/issues/1246 and https://github.com/masayuki0812/c3/issues/1292


aendrew:
LGTM, :+1: — have closed #1251 because this is the better implementation. 

Could you please remove `c3.js` and `c3.min.js` from your commit and rebase? This helps prevent merge conflicts. Thanks!

j-schumann:
+1
Having the same problem, sometimes the subchart x axis is visible as a thin line below the chart since @1168b69
Fix looks good.

Side note: why is the subchart (axis) even generated when the subchart is not shown? Looks like quite a bit of overhead because it is updated each time the chart is zoomed/data is loaded etc.

masayuki0812:
LGTM. Thank you very much!!!

aendrew:
Hoorah! Thanks everyone! Really excited this is fixed now! :+1: :shipit:

Issue #1079, opened 2015-03-25
Overlapping bars


ghost:
Sometimes the bars of a bar chart can overlap:

![screen shot 2015-03-25 at 13 35 14](https://cloud.githubusercontent.com/assets/7317426/6825631/e7a49392-d2f3-11e4-9859-af07b0ef3e9a.png)

**JSFiddle where this can be reproduced on all the major browsers:**
- http://jsfiddle.net/27g7xzwb/

**Steps to reproduce:**
1. Double click an item on the legend, effectively disabling and re-enabling one of the columns before the animation has time to complete.

Whilst the above screenshot is not too dramatic, in practice the overlapping bars can result in completely hidden data, as demonstrated by the two images below. (The light grey bar is hidden behind the dark grey bar.)

![screen shot 2015-03-25 at 13 56 42](https://cloud.githubusercontent.com/assets/7317426/6826209/af9698e8-d2f7-11e4-8444-139836b6e700.png) vs. ![screen shot 2015-03-25 at 13 56 15](https://cloud.githubusercontent.com/assets/7317426/6826208/af914a46-d2f7-11e4-8970-34c12e4822cd.png)

ergo:
This also happens with 0.4.10, when x axis in timeseries data has low amount of ticks, then the bar width is not calculated correctly. When forcing it from ratio to pixel value, then the charts look correct.

panthony:
Cannot reproduce using provided fiddle & current version. Closing.

PR #1324, opened 2015-08-17
Fixed redraw animation bug when user quickly hide and show chart item

Kreozot:
(Issue #1079)

Overlapping happens because of two parallel transitions, on hiding and showing bars (not only bars actually). 
D3 3.5+ allows you to use [named trainsitions](http://bl.ocks.org/mbostock/5d8039fb983a29e2ad49) to avoid collisions.
Because of fact that both of transition are created in one function (redraw*) I solve it with random generated transition name.


aendrew:
Nicely done. LGTM, :+1:, RTBC, &c.

masayuki0812:
Thank you!!

rmosolgo:
Thanks!

Issue #1327, opened 2015-08-17
An option to set space between bars in bar chart


Kreozot:
It would be great to have this feature.

hanu412:
any update on this?

kt3k:
This feature has been released at v0.4.17.

DRUOCM:
I can't find this feature description in the doc

PR #1328, opened 2015-08-17
Added "bar.space" parameter to set space between bars in bar charts.

Kreozot:
(Issue #1327)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -163,6 +163,7 @@ c3_chart_internal_fn.getDefaultConfig = function () {
         bar_width_ratio: 0.6,
         bar_width_max: undefined,
         bar_zerobased: true,
+        bar_space: 0,
         // area
         area_zerobased: true,
         // pie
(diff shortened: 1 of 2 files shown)


Kreozot:
The parameter is multiplier for bar width. So it should be from 0 to 0.5.

aendrew:
@kreozot This is a terrific feature and I really want to add it, would help some layout problems with bar charts. :+1:

Could you please add a unit test or two demonstrating this feature works? At this point, we're requiring unit tests for any new features on the 0.4.x branch — if you need help with that, drop me a line on [Gitter](http://gitter.im/c3js/c3) and I'll walk you through it. Thanks!

panthony:
I am too interested by this feature so I tried to write some basic tests but it appears to have an unexpected behavior in my opinion.

I tried with something like this:

```javascript
            return {
                size: {
                  width: 500
                },
                data: {
                    columns: [
                        ['data1', 30, 200, 100],
                        ['data2', 50, 20, 10],
                        ['data3', 150, 120, 110]
                    ],
                    type: 'bar'
                },
                bar: {
                    space: 0.5
                }
            }
```

I was expecting the bar's width to shrink to make room for the added space but this is not the case. The bar's width stays the same and the container `.c3-chart-bars` growth.

I have the feeling that this will bring issues at some point.

jashetty:
Great feature! Any updates on when this is going to be merged??

kt3k:
closed in favor of #2146.

PR #1337, opened 2015-08-19
fix data order of tooltip

teichmaj:
fixes issue https://github.com/masayuki0812/c3/issues/1028

screenshot from 2015-08-19 16-33-34

screenshot from 2015-08-19 16-33-55

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -29,15 +29,24 @@ c3_chart_internal_fn.getTooltipContent = function (d, defaultTitleFormat, defaul
         titleFormat = config.tooltip_format_title || defaultTitleFormat,
         nameFormat = config.tooltip_format_name || function (name) { return name; },
         valueFormat = config.tooltip_format_value || defaultValueFormat,
-        text, i, title, value, name, bgcolor;
-    
-    if (config.data_order === 'desc') {
+        text, i, title, value, name, bgcolor,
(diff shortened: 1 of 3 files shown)


teichmaj:
Now also with fixed tooltip order for stacked charts
screenshot from 2015-08-19 18-00-11
screenshot from 2015-08-19 18-00-14

src/tooltip.js
```diff
@@ -29,7 +29,27 @@ c3_chart_internal_fn.getTooltipContent = function (d, defaultTitleFormat, defaul
...
+        text, i, title, value, name, bgcolor,
+	orderAsc = $$.isOrderAsc(), orderDesc = $$.isOrderDesc();
+     
+    if (config.data_groups.length === 0) {
+	d.sort(function(a,b){
+	    return orderAsc ? a.value - b.value : b.value - a.value;
+	});
+    } else {
+	var targets = $$.orderTargets($$.data.targets);
+	var data_targets = targets.map(function (i) {
```
masayuki0812: Can you write like:
```
var ids = $$.orderTargets($$.data.targets).map(function (i) {
```
because it's not target, but ids.

masayuki0812:
Can you fix travis and use space for indent, not tab?

masayuki0812:
Thanks!

PR #1344, opened 2015-08-21
Fix error in IE9 when parsing csv files. 

jfrazier:
IE9 doesn't support response param of xhr callback object. Copy responseText param to response param fixed the issue. Charts now render correctly.


jfrazier:
This solution did not work in some browsers due to strict mode. I will refactor and create another pull request.

PR #1354, opened 2015-08-26
Add a function to allow the min/max values to be formated

EspadaV8:
It would be nice to be able to append/prepend things to the min and max values. Currently the only way to do this is with some JS to inject the text after the gauge graph has been rendered. This adds in a format function similar to the label format function.

I wasn't sure, but with PR do you want the concatenated and minified JS included as well?

### DIFF
--- a/htdocs/samples/chart_gauge.html
+++ b/htdocs/samples/chart_gauge.html
@@ -26,6 +26,9 @@
 //            format: function(value, ratio) {
 //              return value;
 //            },
+//            extents: function (value, isMax) {
+//                return value + '%';
+//            },
 //          show: false // to turn off the min/max labels.
           },
 //          min: 0, // 0 is default, //can handle negative min e.g. vacuum / voltage / current flow / rate of change
--- a/spec/arc-spec.js
+++ b/spec/arc-spec.js
@@ -154,6 +154,43 @@ describe('c3 chart arc', function () {
             // This test has bee updated to make tests pass. @TODO double-check this test is accurate.
             expect(data.attr('d')).toMatch(/M-221.*?,-2\..+A221.*?,221.*? 0 1,1 -68.*?,210.*?L-65.*?,201.*?A211.*?,211.*? 0 1,0 -211.*?,-2.*?Z/);
         });
+
+        it('should update labels use custom text', function() {
+            args = {
+                gauge: {
+                    width: 10,
+                    max: 100,
+                    expand: true,
+                    label: {
+                        extents: function (value, isMax) {
+                            if (isMax) {
+                                return 'Max: ' + value + '%';
+                            }
+
+                            return 'Min: ' + value + '%';
+                        }
+                    }
+                },
+                data: {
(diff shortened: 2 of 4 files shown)


nym:
Seems like overkill to have a custom formatter for the max/min, when the center (current) value already has a formatter. Nevertheless, this would solve my issue of not being able to put a $ before the min+max labels on the Gauge.

nym:
Maybe a good compromise is to use the default formatter unless a custom one is provided.

EspadaV8:
Default formatter as in the `label.format` function? Wouldn't that potentially cause more issues? Where people don't want the gauge labels to be formatted at all but do want the middle one formatted. It would also change existing gauges.

And yeah, it probably is overkill to have the `isMax` parameter, but could possibly useful for some people. As a bit of a strained example

```
min: -100,
max: 100
minmaxformat: function (value, isMax) {
   return (isMax) ? "$" + value : "-$" + value.substr(1);
}
```

Would allow the labels to be `-$100` and `$100` instead of `$-100` and `$100` (true, in this example you could check for the `-` in the value, but that might not be possible in all cases).

jkusa:
:+1: to this feature

aendrew:
@EspadaV8 Looks good to me. Thanks for the documentation! :+1: :100:

Will be happy to add it, but could you please write a unit test? All new features must have at least one test.

Nicely done on not adding the build/minified files, please don't add them.

Thank you!

aendrew:
Also, not entirely convince `gauge.minmaxformat` is the most intuitive API. Maybe `gauge.label.extents`, so it's in the label namespace like the main label?

EspadaV8:
Hi @aendrew sorry for the delay in replying. I'll try and get this updated shortly.

kenbier:
👍 Please merge this, submitter addressed the feedback.

kenbier:
any chance this can get merged? @aendrew @EspadaV8 

I'd be willing to chip in on tests if this is otherwise good to merge.

EspadaV8:
@kenbier I'm not sure if I need to do anything but I'd be happy to fix anything up if it needs it before merging.

kenbier:
@EspadaV8 Would you mind adding tests then? According to the reviewer's comments and labels thats all that is necessary.

> Will be happy to add it, but could you please write a unit test? All new features must have at least one test.

I could really use this feature :)

EspadaV8:
Ah, I completely missed that. I'll see if I can add one in, sure.

kenbier:
@EspadaV8 No worries. Also, you'll be asked to squash your commits and rebase against master after you add the test.

EspadaV8:
Added new test, rebased and squashed. I've no idea if the test is any good but it passes so hopefully it'll do 😄  I just copied the first test and added what I needed.

aendrew:
This looks like it's good to merge. Thanks @EspadaV8! :shipit:

EspadaV8:
You're welcome. Glad it painless 😄

PR #1364, opened 2015-09-03
Update axes names in load API call

mbeloshitsky:
Added posibility to update names in load API call.

BTW Is there reason not update 'colors', 'names', 'classes' properties in flow API call?


Telofy:
Oh, this is very useful. Thanks!

Issue #1258, opened 2015-07-08
change axis when loading new data dynamically


karismatik:
Hey guys,
So I am new to json but I have made some good progress.

So I am loading data into an area-spline with a timeseries as the x axis in the form of %Y-%m-%d as shown below

var chart = c3.generate({
 bindto: '#missed-calls',
 data: {
 x: 'time',
 type:"area-spline", 
 colors: {
 missedcall: '#0000ff'
 },
 url: '<?php echo site_url("/dashboard/recent_data/missed_calls/day")?>',
 mimeType: 'json',
 names:{missedcall:"Missed calls"}
 },
 axis: {
 x: {type: 'timeseries', tick:{format: '%Y-%m-%d'}}
 }
 });

I am then loading new data dynamically via json when there is a change in frequency, say weekly, monthly or daily as shown below

$("#dashboard-data").on("change",function(e)
 {
 missedcall_.load({
 url: '/dashboard/recent_data/missed_calls/',
 mimeType: 'json'
 });
});

The x axis timeseries format hence becomes irrelevant as I need it to change from %Y-%m-%d to %m or something iike that..
I am loading the necessary data from json but I do not know how to effect the change of the x axis on the dynamically loaded data..

Where do I place this code to effect the new axis 

axis: {
 x: {type: 'timeseries', tick:{format: '%Y-%m-%d'}}
 }
 }

Issue #1322, opened 2015-08-17
npm install fails on Node ^0.12 (because of old version of grunt-sass)


Kreozot:
Actually this is node-sass issue. Grunt-sass must be updated for newer version of node-sass dependency.
This error described here: https://github.com/sass/node-sass/issues/653

`"grunt-sass": "^1.0.0"` solves the problem

PR #1365, opened 2015-09-03
Load now refreshes x tick format via xtickformat parameter

### LINKED ISSUE #1223: Update axis tick
I have a timeseries spline chart that has different axis tick count and format based on incoming data. So if incoming data is hourly setting `tick.count = 24, and tick.format = '%H %M'`, and if daily it would change to `tick.count = 7, tick.format = '%a, %e %b'`. 

Is there a way to update/refresh chart's axis data?

azurelogic:
- Load now refreshes x tick format via xtickformat parameter.
- Add JetBrains project dir and grunt files to .gitignore.

This should resolve #1223 and #1258.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ d3.min.js
 components
 build
 .sass-cache
+.idea
+.grunt
+_SpecRunner.html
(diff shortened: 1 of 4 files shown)


azurelogic:
Also, please integrate #1322. I had to recreate those changes just to work on this PR.

azurelogic:
Any thoughts on merging this in?

Issue #215, opened 2014-05-05
ondragstart and ondragend callbacks prevent scrolling on touch devices


sandrolino:
I noticed that c3js graphs disable somewhat scrolling on touch devices, probably due to their interactivity. 
Disabling Tooltip still results in interactivity when hovering over data points. 
Is there away to completely disable hovering over data points (i.e an svg) such that scrolling works on c3js graph works also on touch devices ?

Thanks for your help.
s

moklick:
It would be very helpful if you could add something like swipe detection. By doing this we could interact with the charts even on mobile devices. Right now we have project where we have several charts in an article. Users can't scroll on the charts, so we have to disable interaction which is sad, because interaction works great on mobile devices.

aendrew:
I started a new issue about this recently (since closed) due to having not noticed this one. I'm moving discussion to this thread (and reopening) because there's been more work done here.

Important point from #798: [the problem is in the ondragstart and ondragend callbacks](https://github.com/masayuki0812/c3/issues/798#issuecomment-67310392). Disabling those on ln. 233 in `interaction.js` fixes the scrolling issue while preserving interactivity.

moklick:
@aendrew wow!! thanks you very much. commenting out these lines fixes the problem for us. awesome!

aendrew:
@masayuki0812 Any update on removing those callbacks? Would it help if I did a PR? 

I'm currently integrating [axisJS](http://times.github.io/axisJS) into a Node app that allows C3 charts to be used really effortlessly in news stories; however, it totally causes issues on the iPad due to this issue. I can run off a fork for the moment, but it'd be really nice to be able to get that into core...

masayuki0812:
I added `data.selection.draggable` option that enables dragging to select the points. The default of this option is `false` and then drag interaction will not be added to the chart, so it means we can scroll even if it's on the chart as default.
On the other hand, if we want dragging for data selection, we need to set this option `true`, but I think it's acceptable because the use case seems not much and the scrolling is more important as default interaction.
I believe this fix works for this issue and I'll release the fix in the next version v0.4.9 shortly.

masayuki0812:
v0.4.9 has been released, so please let me close.

aendrew:
Yay! :+1: :shipit: Nicely done, @masayuki0812!

aendrew:
Adding `documentation` tag re: `data.selection.draggable`.

moklick:
nice! thanks @masayuki0812

[...]

PR #1367, opened 2015-09-03
On drag events

### LINKED ISSUE #1366: Add ondragstart and ondragend events
In our application we would like to unselect the current selection in the chart when the drag start and trigger an event unrelated to C3 when the drag end. From the code, only the events seems to be missing

GermainBergeron:
fixes #1366


borisnadion:
any insights when this will be merged?

aendrew:
It doesn't seem like there's much interest in either this or #1366. If you provide a better use case and/or there's more appetite for it, I'll reopen. 

Closing for now.

unstephenk:
Would love to have this feature.

IceCreamYou:
It appears this code used to be part of the library but was removed in #215. Here is a workaround:

```javascript
// Override C3's dragend function so we can do something when the user is done selecting multiple points
const dragend = c3.chart.internal.fn.dragend;
c3.chart.internal.fn.dragend = function() {
    dragend.apply(this, arguments);
    // Do whatever you want to do on dragend here
};
```

PR #1382, opened 2015-09-14
Upgrade d3 to 3.5.4

bouk:
@masayuki0812 I believe d3 can be upgraded to 3.5.4, as https://github.com/mbostock/d3/issues/2174 has been closed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "gitHead": "84e03109d9a590f9c8ef687c03d751f666080c6f",
   "readmeFilename": "README.md",
   "dependencies": {
-    "d3": "<=3.5.0"
+    "d3": "^3.5.4"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


thomblake:
This would be super-helpful; it looks like this is the main thing blocking us from upgrading to Node 4.0 (since newer d3 no longer depends on jsdom).

bouk:
Correct, I'm using my branch of c3 with node v4 for now

aendrew:
Good to know re: 3.5.4 and NodeJS, though closing as duplicate because we're discussing this over at #1166. 

Regardless, I'd rather it be `~3.5.4` instead of `^3.5.4` as that will constrain it to bugfix releases (and not minor releases; upgrading D3 from 3.5.x to 3.6.x will likely cause problems we'll need to handle).

PR #1390, opened 2015-09-22
Add .scss source files to bower package

MarcBT:
Hi !
It can be useful to have access to .scss files in the bower package for some cases.
As an exemple, it can simplify a gulp / grunt / brunch task handling styles when all the bower components of the project expose their .scss source files.
Regards.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -20,7 +20,7 @@
     "bower_components",
     "htdocs",
     "spec",
-    "src",
+    "src/**/*.js",
     "package.json",
     "component.json",
     "Gruntfile.*"


aendrew:
@MarcBT Hi!

I don't disagree at all with adding .scss to bower.json, but that's not at all what your commit does — it simply adds `src/**/*.js` to bower.json's ignore. This actually removes stuff from `bower` instead of adds it...

If you amend your PR to in fact include the Sass files, I'll RTBC. Thanks!

MarcBT:
From bower doc, Bower ignore follows the same rule as gitignore pattern spec.
If you prefer, I can change the PR to make it match the latest exemple provided in [gitignore documenation](http://git-scm.com/docs/gitignore#_examples)

aendrew:
@MarcBT If you want to make it a bit easier to initially understand, maybe like:

```
'src/',
'!src/scss/`
```

Or something like that? Regardless, it doesn't really matter, your PR's perfectly fine!

MarcBT:
All right ! So let's maintain like this so that no need to amend the commit. 
Thanks @aendrew.

kt3k:
Thanks for the contribution!

PR #1395, opened 2015-09-25
More accurate positioning of y axis tick texts for larger font sizes

900k:
While using c3js in a project where people demanded larger font sizes for axis ticks, I noticed that changing the font size to 12pt is causing the y axis tick texts to be vertically out of position relative to the actual ticks. The tick texts are not vertically centred at the ticks any more. Here is an example for 12pt:
12pt_ugly_
This problem becomes more obvious if you increase the font size. Here an example for 20pt:
20pt_ugly
The changes applied by me lead to the following results for 12pt and 20pt:
12pt
20pt
Also tried other font sizes from 6pt to 30pt and the results are still accurate.
I've found the reason for this problem in the methods tspanDy(d, i) and textFormatted(d). In tspanDy(d, i), a fixed value (3) has been used to centre the tick text vertically, but using a quarter of the height of a char to compute the vertical offset of tick texts results in a more accurate positioning. The method textFormatted(d) is used to detect the height of the chars in the tick texts. But it fails if the tick "text" is a number. To solve this issue I ensured that textFormatted(d) will return a string in any case.

Do you think these changes are ok?

### DIFF
--- a/src/c3.axis.js
+++ b/src/c3.axis.js
@@ -151,7 +151,7 @@ function c3_axis(d3, params) {
                 var dy = sizeFor1Char.h;
                 if (i === 0) {
                     if (orient === 'left' || orient === 'right') {
-                        dy = -((counts[d.index] - 1) * (sizeFor1Char.h / 2) - 3);
+                        dy = -((counts[d.index] - 1) * (sizeFor1Char.h / 2) - sizeFor1Char.h / 4);
                     } else {
                         dy = ".71em";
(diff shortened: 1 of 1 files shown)


900k:
Ok, just noticed that the integration tests failed. Will have a look into that. Is there a way to perform the tests in my repository so that I don't disturb you here?

900k:
I will close this request and open an issue first.

PR #1400, opened 2015-09-30
Added api axis.cullingMax

### LINKED ISSUE #1356: Please allow updating culling value after load
It is now impossible to change the culling value after the initial load. This is an important feature for a scenario when you have lots of ticks and you want to show/hide different amounts of them when zooming in and out.

Thanks!

parthapal33:
Added new api axis.cullingMax to update the max tick culling value after chart is loaded. This pull request fixes #1356.

### DIFF
--- a/spec/api.axis-spec.js
+++ b/spec/api.axis-spec.js
@@ -51,4 +51,28 @@ describe('c3 api axis', function () {
         });
 
     });
+
+	describe('axis.cullingMax',function(){
+		it('should set max culling',function(){
+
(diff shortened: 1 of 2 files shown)


z:
This looks good.

neothemachine:
Just tried to use this inside the `zoomend` event handler and I see a weird behaviour. I can zoom in, but it doesn't let me zoom out again. When I remove the call to `cullingMax(...)` the zoom-out works again. Any idea what could cause this?

Issue #1399, opened 2015-09-29
Issues installing on Windows


bartvde:
I'm having issues installing on Windows. Error messages such as:

![screen shot 2015-09-29 at 20 28 42](https://cloud.githubusercontent.com/assets/319678/10174177/bf8563bc-66e8-11e5-9c49-35757d59c08b.png)
![screen shot 2015-09-29 at 20 28 35](https://cloud.githubusercontent.com/assets/319678/10174178/bfa7f922-66e8-11e5-8d15-9d6b8abb2bda.png)

I also have d3 as a dependency in my project and 3.5.6 builds fine on Windows. Is there a specific reason c3 is tagged to older version of d3?

Fox32:
Just run into the same problem, any chance a higher version could be used?

Issue #1392, opened 2015-09-23
Are there known issues with node 4.1.1 ?


bartvde:
```
Barts-MacBook-Air:sdk bartvandeneijnden$ npm i c3

> contextify@0.1.14 install /Users/bartvandeneijnden/opengeo/git/sdk/node_modules/c3/node_modules/d3/node_modules/jsdom/node_modules/contextify
> node-gyp rebuild

  CXX(target) Release/obj.target/contextify/src/contextify.o
In file included from ../src/contextify.cc:3:
../node_modules/nan/nan.h:261:25: error: redefinition of '_NanEnsureLocal'
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
                        ^
../node_modules/nan/nan.h:256:25: note: previous definition is here
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Handle<T> val) {
                        ^
../node_modules/nan/nan.h:661:13: error: no member named 'smalloc' in namespace 'node'
    , node::smalloc::FreeCallback callback
      ~~~~~~^
../node_modules/nan/nan.h:672:12: error: no matching function for call to 'New'
    return node::Buffer::New(v8::Isolate::GetCurrent(), data, size);
           ^~~~~~~~~~~~~~~~~
/Users/bartvandeneijnden/.node-gyp/4.1.1/include/node/node_buffer.h:31:40: note: candidate function not viable: no known conversion from 'uint32_t' (aka 'unsigned int') to 'enum encoding' for 3rd
      argument
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
                                       ^
/Users/bartvandeneijnden/.node-gyp/4.1.1/include/node/node_buffer.h:43:40: note: candidate function not viable: 2nd argument ('const char *') would lose const qualifier
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
                                       ^
/Users/bartvandeneijnden/.node-gyp/4.1.1/include/node/node_buffer.h:28:40: note: candidate function not viable: requires 2 arguments, but 3 were provided
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
                                       ^
/Users/bartvandeneijnden/.node-gyp/4.1.1/include/node/node_buffer.h:36:40: note: candidate function not viable: requires 5 arguments, but 3 were provided
[...]

PR #1414, opened 2015-10-07
re #1392 #1399 Update package.json to use D3 v3.5.6 for Windows issue

kahboom:
This PR fixes an issue for Windows users that want to use C3. C3 depends on D3, which depends on JSDOM, which depends on Contextify, which uses node-gyp, which causes issues on Windows due to a lack of a C/C++ compiler, the need to install Python, etc. D3 has made JSDOM an optional dependency in its most recent version, though C3 is still automatically installing D3 v3.5.0. So, any user on Windows that is using C3 will run into these issues unless the version is changed in the `package.json`.

Ref: https://github.com/mbostock/d3/issues/1986

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "gitHead": "84e03109d9a590f9c8ef687c03d751f666080c6f",
   "readmeFilename": "README.md",
   "dependencies": {
-    "d3": "<=3.5.0"
+    "d3": "3.5.6"
   },
   "devDependencies": {
     "grunt": "^0.4.5",


bartvde:
Fixes #1399 as well

kahboom:
Thanks @bartvde , good call.

ChuckPierce:
This is such an easy fix that will make the c3 package compatible with new versions of node and npm. Would be great if this could get merged in!

aendrew:
This is irrelevant as of cc8dae366357a20af3e64df1aff92cfa7c0abb1d. NPM should now install the latest D3 in the 3.5.x branch. Closing.

bartvde:
thanks, when will this be published to npm?

aendrew:
@bartvde Either the next RC for 0.4.11, or 0.4.11 itself — whichever's next. Open issues on that milestone are [here](https://github.com/masayuki0812/c3/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.4.11).

PR #1419, opened 2015-10-09
Duplicate key

vlakarados:
Key defined below, has different value. Keeping second one as it overwrites the first one anyway.


kt3k:
Good catch! Thanks!

PR #1426, opened 2015-10-11
Updated grunt jasmine to use the CDN version of D3. Fixes #1425.

### LINKED ISSUE #1425: Serving D3 via the non-production Rawgit URL is leading to test failures
I'm working with local copy of latest c3 and want to test it.
When I run `grunt` I get a lot of errors `TypeError: 'undefined' is not an object (evaluating 'd3.scale') in file:///srv/missingdays/Projects/c3/c3.js (line 5708) (1)` on this line

``` js
pattern = notEmpty(config.color_pattern) ? config.color_pattern : d3.scale.category10().range(),
```

I don't seem to find any other way to test that library is actually installed correctly.
P.S. Yes, I know I can use examples but this is not the solution I want/

aendrew:
This should definitely use the CDN version; it's possible tests are failing due to hitting the usage limit. See #1425.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -92,7 +92,7 @@ module.exports = (grunt) ->
               specs: 'spec/*-spec.js'
               helpers: 'spec/*-helper.js'
               styles: 'c3.css'
-              vendor: 'https://rawgit.com/mbostock/d3/v3.5.0/d3.min.js'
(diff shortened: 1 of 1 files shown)


kt3k:
@aendrew 
As you already know, we switched to karma and this isn't a problem any more. Thanks for the contribution anyway! :)

PR #1451, opened 2015-10-30
Add footer config to c3 charts

davidpoore:
@aendrew First pass at https://github.com/masayuki0812/c3/issues/1266 - added a footer config similar to the header config in my other pull request. Also added ability to send the title to the bottom of the chart, which necessitated some sizing changes.


aendrew:
I'm going to put this `onhold` for the moment because as per #1032, this probably needs to be a plugin. Luckily, we're hoping to start making progress on that in the next month or two, so stay tuned. 😄

Braintelligence:
@aendrew Anything new about this functionality? :)

PR #1457, opened 2015-11-06
change d3 version, so that npm install doesn't install contextify

juliankrispel:
Contextify, a dependency of jsdom, which in turn has been a dependency of d3 in version 3.5, still gets installed when I go `npm install c3@latest`. 

This breaks c3 if you install it via node4 (because building contextify in node 4 isn't possible right now)

Upgrading to the latest minor version of d3 fixes this, because jsdom has been made a dev dependency rather than a hard dependency.


aendrew:
Duplicate of #1166.

PR #1462, opened 2015-11-09
fix auto bar width on timeseries bar chart

gwokae:
[issue demo @ jsfiddle](http://jsfiddle.net/gwokae/chfd9wdp/2/)

When the width of bar chart is not given,  0.4.11-rc\* will auto calc a with for it. It works on non-timeseries charts. But timeseries bar chart will overlapped when the x value not distributed evenly.


shvelo:
+1 affected by this issue

jozefvaclavik:
👍  having same issue.

panthony:
Thanks for the contribution this suggested fix seems more complex than necessary with no tests attached.

Since this PR was opened 3 years ago I'll close it since I doubt you'll want to continue.

Issue #1465, opened 2015-11-12
SvgPathSeg is removed in chrome 48


Karabur:
According to https://www.chromestatus.com/features#pathseg this interface was a part of svg1.1, but was removed.
Soon charts will be broken in chrome. 
And perhaps it will be removed from other browsers too. 
Library should use other way to avoid reference to pathSegList property.

aendrew:
Good catch, @Karabur!

This needs to be updated in the following places:

- [ ] [src/util.js][1]
- [ ] [src/shape.bar.js][2]

[1]: https://github.com/masayuki0812/c3/blob/931109f8269688b8d623b4523335670dd813b4dc/src/util.js#L43
[2]: https://github.com/masayuki0812/c3/blob/14b7f4ca4e91094d16458a83c5ce654c08287361/src/shape.bar.js#L117

martingraham:
Could this polyfill be bundled into c3? --> https://github.com/progers/pathseg

I suppose we could also add it individually to our projects in the meantime

aendrew:
@martingraham Ugh, would really like to not bundle another polyfill unless totally necessary. SvgPathSeg is only used three times in the entire codebase, would prefer just finding an alternative implementation.

GerHobbelt:
No polyfill required; the SVG path list is only used in a context where a straight bar block (rectangle) is expected (segment 0 and segment 1 are inspected in there to derive a corner coordinate); the above pullreq code keeps the existing code for those who still have SVG pathSegList support, while the alternative code works or Chrome Canary et al - at least it works for me. ;-)

aendrew:
@GerHobbelt You, good sir, are a gentleman and a scholar. :+1: :tada: 

Will review PR first opportunity. Many thanks!

aendrew:
Apologies for the delay, will try to get a chance to look at this this week.

GerHobbelt:
No worries. ;-)

tracker-common:
Hey @aendrew, just wanted to touch base and see if this bug is still in progress.

Thanks!

bsarkadi:
Hi @aendrew, I'd also like to ask when the fix for this is expected to be released. Thanks!

Paulsky:
Hello,

[...]

PR #1466, opened 2015-11-16
extracted out of mainline: patch for #1465

GerHobbelt:
origin: commit SHA-1: b7ed4407467fba164d87127ec5e31ef4d20e28d6
- fix crash in click handler due to use of pathSegList property which is not available apparently in Chrome at least: now the code is cleaned up by using the internal `getPathBox()` function, where the code is corrected to cope with this issue by taking the coordinates straight off the calculated SVG box.

_Notes_ See also: https://github.com/masayuki0812/c3/issues/1465 (#1465)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -40,7 +40,22 @@ var isValue = c3_chart_internal_fn.isValue = function (v) {
     },
     getPathBox = c3_chart_internal_fn.getPathBox = function (path) {
         var box = path.getBoundingClientRect(),
-            items = [path.pathSegList.getItem(0), path.pathSegList.getItem(1)],
-            minX = items[0].x, minY = Math.min(items[0].y, items[1].y);
-        return {x: minX, y: minY, width: box.width, height: box.height};
+            minX, minY;
+        // MSIE supports pathSegList while it crashes on latest Chrome: https://msdn.microsoft.com/en-us/library/ff971976(v=vs.85).aspx
+        // See also: https://github.com/masayuki0812/c3/issues/1465
+        if (path.pathSegList && path.pathSegList.getItem) {
+            var seg0 = path.pathSegList.getItem(0);
+            var seg1 = path.pathSegList.getItem(1);
+            minX = Math.min(seg0.x, seg1.x); 
+            minY = Math.min(seg0.y, seg1.y);
+        } else {
+            minX = box.left;
+            minY = box.top;
+        }
+        return {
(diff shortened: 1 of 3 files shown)


src/util.js
```diff
@@ -40,7 +40,22 @@ var isValue = c3_chart_internal_fn.isValue = function (v) {
     },
     getPathBox = c3_chart_internal_fn.getPathBox = function (path) {
         var box = path.getBoundingClientRect(),
-            items = [path.pathSegList.getItem(0), path.pathSegList.getItem(1)],
-            minX = items[0].x, minY = Math.min(items[0].y, items[1].y);
-        return {x: minX, y: minY, width: box.width, height: box.height};
+            minX, minY;
+        // MSIE supports pathSegList while it crashes on latest Chrome: https://msdn.microsoft.com/en-us/library/ff971976(v=vs.85).aspx
+        // See also: https://github.com/masayuki0812/c3/issues/1465
+        if (path.pathSegList && path.pathSegList.getItem) {
```
danevans: What's the trade off of simplifying this so that it doesn't include any references to `pathSegList` and just calls `box.left` and `box.top`? That seems like a cleaner approach because there would be only one code path.

src/util.js
```diff
@@ -40,7 +40,22 @@ var isValue = c3_chart_internal_fn.isValue = function (v) {
...
+            var seg0 = path.pathSegList.getItem(0);
+            var seg1 = path.pathSegList.getItem(1);
+            minX = Math.min(seg0.x, seg1.x); 
+            minY = Math.min(seg0.y, seg1.y);
+        } else {
+            minX = box.left;
```
danevans: Testing this with a breakpoint at line 1364 inside the `isWithinBar` function, on [one of the demo charts](http://c3js.org/samples/chart_combination.html) it appears that `box.left` is not the same as `seg0.x`.

screen shot 2016-02-16 at 11 35 31 am

I've seen some people [subtract the svg container's `boundingClientRect` from the element's](http://stackoverflow.com/a/18554375) but I'm not sure if that's what we want here because those results aren't exactly equal to the `pathSegList` values either.

screen shot 2016-02-16 at 11 42 02 am

aendrew:
@GerHobbelt Mind addressing @danevans' comments? I'd like to eventually update things so that we totally pull out all the PathSegList code and remove the polyfill we presently have, such as this PR does. Thanks!

PR #1505, opened 2015-12-13
Implemented access to nested JSON objects as values for charts

kelthuzad:
This is the code to implement https://github.com/masayuki0812/c3/issues/1471

Code to access objects by path was taken from the solution of
https://stackoverflow.com/questions/6491463

When JSON objects are used to load chart data, nested / inner objects and
arrays can now be loaded, too.
See additional tests in spec/data-spec.js for examples, how to access nested
JSON object values


kelthuzad:
Checks fail because I did not commit the changes in c3.js and c3.min.js.
Will open a new request with the fixed version.

Issue #508, opened 2014-08-19
I can't put the legend on the left


JulianMLaf:
I have to do something like this http://bl.ocks.org/mbostock/3888852 

And in the first donut I can't put the legend on the left

jsfiddle: http://jsfiddle.net/aZLJ8/3/

Thank you, Julián.

PR #1521, opened 2015-12-30
New legend position options: "left" and "top"

GoekhanOezen:
#508

Hi,
In our project we need to position legend left and top so I tried to enchanch C3 for this prupose. The generated charts have the exact same height and width for `legend_position` `left` and `top` as they have for `legend_position` `right` and `bottom`, even for arcs.
We will use this anyway but we want to share all our enchanchments and hope at least some of them will be accepted :-)

legend-left-and-right

### DIFF
--- a/spec/legend-spec.js
+++ b/spec/legend-spec.js
@@ -140,6 +140,59 @@ describe('c3 chart legend', function () {
 
     });
 
+    describe('legend as top', function () {
+
+        it('should change the legend to "top" successfully', function () {
+            args = {
+                data: {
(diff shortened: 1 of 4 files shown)


GoekhanOezen:
It seems you check the added test with the old C3.js (without the committed changes of mine)

cpcarrot:
The ability to position the legend at the top would be very useful for me. Any further progress on incorporating this improvement into a release?

mpires-dev (approved):

PR #1533, opened 2016-01-10
This closes #1531 Legend double click hide all but the clicked

### LINKED ISSUE #1531: Legend double click hide all but the clicked
When legend is clicked twice all legends are hide, but the clicked.

http://jsfiddle.net/su2axgsd/4

souzadevinicius:


panthony:
This feature already exists using 'alt + click'.

If at one point a dblckick callback is added maybe it should as a config to allow the user to do what he wants with this.

PR #1550, opened 2016-01-21
fixed isWithinBar() crash when there were less than 2 items in pathSegList

adufilie:

PR #1551, opened 2016-01-21
fixed getPathBox() crash when there were less than 2 items in pathSegList

adufilie:


aendrew:
@adufilie Thanks for all these PRs, trying to merge a few of them. Would you mind perhaps writing a test for this and #1550? Also, what's the particular error case that triggers the error described in the title? Thanks!

PR #1552, opened 2016-01-22
Removed use of SvgPathSeg, fixing Chrome error

bottomline:


perlun:
@masayuki0812 - any chance we can get this merged (and a new version released) in the near future? The current version is broken in Chrome 48, which was rolled out some week ago, so it's quite critical to get this fixed.

(For now, using the version from bottomline, but it would be nice to be able to get this into an "official" 0.4.11 release or similar.)

Thanks in advance, and huge thanks for a _great_ FOSS project! Do let us know if this is a time problem; I'm sure there are volunteers that can help with maintaining the project if needed.

ericpias:
The code change in this pull request uses a reserved javascript keyword (char) as a variable name in several spots.  In testing it with yuicompressor, it fails.  It is a simple matter to change it.  When I get more comfortable with github I will change it if not already fixed.  I would hate for it to get into a released version that way.  Thanks for for fixing the issue!

src/util.js
```diff
@@ -1,3 +1,312 @@
+// PathData object is based upon `Source`, here https://github.com/jarek-foksa/path-data-polyfill.js/blob/master/path-data-polyfill.js
```
LoneRifle: The nature of the changes in this file strike me as a polyfill. I'm not an active contributor to this project, but as an interested consumer, I would think it might make more sense if this were moved to src/polyfill.js.

For what it's worth, I favour this PR as a whole over #1564, as we should be providing forwardbridges to current APIs, not backbridges to legacy ones.

SSKSoftware:
Downloaded the latest zip from https://github.com/masayuki0812/c3
c3 version : 0.4.11-rc4
d3 version : d3-3.5.0.
and using crome version : Version 48.0.2564.97 m

but still the click event fails.
:Uncaught TypeError: Cannot read property 'getItem' of undefined

ggamir:
Please... everyone's browsers are autoupdating to the latest version... it's only a matter of time... :cry:

perlun:
I agree, it's becoming a real issue. Is the project abandoned?

aendrew:
I've merged #1564 into `dev` so am closing this as a duplicate.

PR #1554, opened 2016-01-22
firefox takes much time to concat a big array

hn-nguyen:


aendrew:
Minor fix, but good spot. Tests aren't really needed, it's one line. Am marking as RTBC, thanks for your PR!

Issue #1235, opened 2015-06-24
Uncaught TypeError: Cannot read property 'x' of null


mmiszy:
if you call `chart.flow(…)` twice immediately or after a short period of time you get the following error:
> Uncaught TypeError: Cannot read property 'x' of null

It's thrown in `c3.js` at line `6227`:
```javascript
            } else if (flow.orgDataCount === 1 || flowStart.x === flowEnd.x) {
```
because `flowStart` is `null`.

Here's a CodePen illustrating the problem: http://codepen.io/mmiszy/pen/xGPYEj
The problem might occur even if the second call to `chart.flow(…)` happens after 300ms (maybe longer?)

sushitommy:
I have the same problem!

mmiszy:
Any info on this one @masayuki0812 ? We get hundreds of these exceptions in our live graphs.

abhas9:
Having same problem in my code. Any insight on this from other developers would be really helpful. Thanks in advance.

vim-zz:
same here

jakerella:
I see the same. It looks like the `flowStart` variable may be null if a flow is started when there is no data:

`flowStart = $$.getValueOnIndex($$.data.targets[0].values, flowIndex),`

In that line, the `values` will be null on a new chart (no data yet). For me, if I seed the chart with a single data point, then start the flow afterward, the issue is resolved.

One solution may be to simply check that variable in [the `if` condition](https://github.com/masayuki0812/c3/blob/master/src/api.flow.js#L205):

```js
  ...
  } else if (flow.orgDataCount === 1 || (flowStart && flowStart.x) === (flowEnd && flowEnd.x)) {
  ...
```

This worked for me anyway... I'll submit a PR for it.

myfbone:
Have the same problem now
```
Uncaught TypeError: Cannot read property 'x' of null
    at c3.js:4571
    at ChartInternal.redraw (c3.js:1822)
    at Chart.flow (c3.js:4494)
```

PR #1576, opened 2016-02-11
Fixes issue of trying to flow data when there is no data to start

jakerella:
This PR fixes issue #1235 where errors are seen in the console when trying to flow data to a chart that does not yet have any data in it. The fix merely checks that the flow start and flow end have values before checking their respective `x` properties (which resulted in a null property check otherwise).


aendrew:
Awesome, many thanks! Merging into `dev`!

jakerella:
np, I'm happy to submit the quick fixes. ;)

masayuki0812:

PR #1577, opened 2016-02-14
Allow tooltip formatter to change output depending on all data points.

SSchwaiger:
I often need to adjust the values presented to the user in the tooltip. Therefore I need information about all data points being displayed for the current point in time.

BTW: I mainly need this because I cannot format the contents of a bar chart depending on its sign.


aendrew:
Merged into `dev`. Not requesting tests because it's a minor change. Thanks for your PR!

masayuki0812:

PR #1578, opened 2016-02-15
dont order items in tooltip if order value in config is set to null

LukasJenicek:


aendrew:
Hi @LukasJenicek!

Nice PR! Could you please do the following?

1. Add a unit test demonstrating the changed functionality works as expected. This helps us ensure your changes work, as well as prevents other changes from breaking yours. Thanks!
2. Please also remove c3.js and c3.min.js from your commit as these get automatically built during releases.

Thank you!

panthony:
This was later fixed, and even a new `tooltip_order` was added to have a different sorting between data and tooltip.

PR #1582, opened 2016-02-17
replace drag selection XOR with more reasonable behavior.

adufilie:


aendrew:
@adufilie Could you please describe what the current behaviour is and how this PR changes that? Thank you!

adufilie:
You can see the problem if you add
`    selection: { enabled: true, multiple: true, draggable: true },`
to the "data" config in the example here: http://c3js.org/samples/chart_scatter.html

The current behavior performs an XOR selection operation on shapes falling within the selection rectangle. The logic is flawed - if you move the mouse fast, the selection becomes incorrect.  If the user attempts to correct the selection, it becomes very difficult to do with the XOR behavior.  This patch makes it so it replaces the selection with the shapes that fall within the most recent selection rectangle.

PR #1595, opened 2016-02-23
Adds config option in area hash to draw the area fill above the line …

linearza:
Adds config option in area hash to draw the area fill above the line instead of below.

Related issue: https://github.com/masayuki0812/c3/issues/1594


aendrew:
Wow, that's so cool! Merging into `dev`! :tada:

masayuki0812:
This has been released as 0.4.11. Thanks!

PR #1610, opened 2016-03-01
point.select.r can now can accept a function

adufilie:
Previously, point.select.r was defined by either a single constant value or would default to 4 times the radius.  Now it can be defined by a custom function.


lbayerl:
+ 1

aendrew:
Nicely done (and good idea, I've wanted that myself)! I've merged this into `dev`.

masayuki0812:

PR #1625, opened 2016-03-18
Implemented access to nested JSON objects as values for charts

kelthuzad:
This is the code to implement https://github.com/masayuki0812/c3/issues/1471

Code to access objects by path was taken from the solution of
https://stackoverflow.com/questions/6491463

When JSON objects are used to load chart data, nested / inner objects and
arrays can now be loaded, too.
See additional tests in spec/data-spec.js for examples, how to access nested
JSON object values

I also changed the negative precision of some tests because I think they didn't make sense.


aendrew:
Lovely PR! I've merged it into `dev`.

masayuki0812:
This has been released in 0.4.11. Thanks!

kelthuzad:
Nice, glad I could help!

PR #1636, opened 2016-03-30
Fixes error that causes permanently fail of chart creation

bnest:
in some cases d3.event.sourceEvent is null (I don't really know why)... In my case that happens when i change thw window soze of the browser window and then zoom in charts with my infrared multitouch TV, then a exception is thrown "can't read the property of type on 'null' ". After that error is thrown (I also don't know why) its not possible to create new charts with c3.

this will fix the issue


aendrew:
Hi @bwestpha!

Seems like a sensible fix to me; mind putting your changes in `src/zoom.js` instead (and removing c3.js from your commit), and then amending your PR? c3.js gets overwritten on `grunt concat`.

Thanks!

PR #1641, opened 2016-04-01
Fixes some tick are missing to display when chart type is category and categories array is supplied

Dekai:
It's related to issue https://github.com/masayuki0812/c3/issues/1638;
Basically it caused by method Axis.prototype.generateTickValues and c3_chart_internal_fn.categoryName.

The tick value generated by Axis.prototype.generateTickValues, the decimal value passed in as index of array config.axis_x_categories[i], it can't get the categoryName, just return undefined.

```
 c3_chart_internal_fn.categoryName = function (i) {
        var config = this.config;
        return i < config.axis_x_categories.length ? config.axis_x_categories[i] : i;
    };
```

Fixing it by adding categoryIndex = Math.ceil(i);

c3.chart.internal.fn.categoryName = function (i) {
    var config = this.config, categoryIndex = Math.ceil(i);
    return i < config.axis_x_categories.length ? config.axis_x_categories[categoryIndex] : i;
};
If we fix it inside generateTickValues to generate integer tick value , it will cause splitTickText doesn't work properly.
From:
`tickValues.push(forTimeSeries ? new Date(tickValue) : tickValue);`
To:
`tickValues.push(forTimeSeries ? new Date(tickValue) : $$.isCategorized() ? Math.ceil(tickValue): tickValue);`

### DIFF
--- a/src/category.js
+++ b/src/category.js
@@ -1,4 +1,4 @@
 c3_chart_internal_fn.categoryName = function (i) {
-    var config = this.config;
-    return i < config.axis_x_categories.length ? config.axis_x_categories[i] : i;
+    var config = this.config, categoryIndex = Math.ceil(i);
+    return i < config.axis_x_categories.length ? config.axis_x_categories[categoryIndex] : i;
 };


martinpovolny:
This seems to work for us. See screenshot in https://github.com/ManageIQ/manageiq/pull/8347. Can this be merged, please?

aendrew:
@dekai I'd love the merge this, but you've made the changes to c3.js, not `src/category.js` — this means your changes will get overwritten on build!

Could you please move them there and update your PR? Please remove c3.js from your commit as well as it gets generated via `grunt concat`. Thanks!

martinpovolny:
@Dekai : if you have problems doing it as @aendrew suggested, ping me, we will get it done. Thx!

Dekai:
@martinpovolny I have updated src/category.js, and reverted changes of c3.js. But c3.js file still in the commit because of minor format different. If possible, please check on it. Thanks.

skateman:
Try it like this:
```bash
git fetch https://github.com/skateman/c3.git dekai-fix:dekai-fix
git push -f origin dekai-fix:master
```
Please let me know if it is done so I can delete my branch.

Dekai:
@skateman Done with your command, thanks!

panthony:
Shouldn't `categoryIndex` be compared against `config.axis_x_categories.length` instead of `i` ?

Let's say `i=1.2` and `categories=[a,b]`, the condition `1.2 < 2` would be true whereas the index 2 would return undefined.

kt3k:
I agree with @panthony .

I think the second line should be:
```
return categoryIndex < config.axis_x_categories.length ? config.axis_x_categories[categoryIndex] : i;
```

Issue #1670, opened 2016-04-28
Centered labels option


qwelias:
![expected](http://i.imgur.com/X2pZ8pc.png)

PR #1671, opened 2016-04-28
Centeredlabels initial

qwemaze:
#1670


qwemaze:
I'm a little bit struggling with locating a correct place to land check if there's enough space for label.

PR #1672, opened 2016-04-29
Add bar with max width example

rodrigowirth:


aendrew:
Looks good, but could you please do the following?

- [ ] Remove the extraneous commented lines in the example.
- [ ] Remove the extraneous comma after `bar.width`

Thanks!

rodrigowirth:
@aendrew, I've just applied both suggestions.

PR #1679, opened 2016-05-02
Updated .travis.yml to do a full build

aendrew:
The current Travis config won't ever pass newly-added unit tests because it doesn't do an actual build from the updated source files. This just replaces the `lint/test` lines with `grunt`, which runs linting and testing anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ before_script:
 - gem install sass
 
 script:
-- npm run lint
-- npm test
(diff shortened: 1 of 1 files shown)


aendrew:
I'm quickly merging this into `master` because not doing so is dangerous (tests will pass for new code that might be broken, because it's using a known-working version of C3 and not the new one) and hinders the work of anyone who wants to do a PR.

Issue #1691, opened 2016-05-10
Unable to create chart with json and dot in key


DoomyTheFroomy:
I have tried to create a `donut` chart for Chrome version. Which contain a dot as key. This will cause errors. I have tried to create the chart on [http://c3js.org](http://c3js.org/samples/chart_donut.html) ([c3.min-4c5bef8f.js](http://c3js.org/js/c3.min-4c5bef8f.js)) with a chrome (50.0). But it was not possible. 
<img width="395" alt="bildschirmfoto 2016-05-10 um 13 43 58" src="https://cloud.githubusercontent.com/assets/1259935/15145350/482f45d0-16b5-11e6-8f86-e0be3e2e0f62.png">

``` js
var chart = c3.generate({
    data: {
        json: [
            {'38.0': 10, '46.0': 10, '47.0': 10, '49.0': 20, '50.0': 30, 'unbekannt': 4}
        ],
        keys: {
            value: ['38.0', '46.0', '47.0', '49.0', '50.0', 'unbekannt']
        },
        type : 'donut'
    },
    donut: {
        title: "Chrome Versions"
    }
});
```

Kyhel:
I encountered the same problem today.
I've done some research and found the culprit code.

In the function **c3_chart_internal_fn.convertJsonToData**, since the version 0.4.11 (the problem doesn't exist in version 0.4.10), the following block of code changes :

##### 0.4.10
```javascript
targetKeys.forEach(function (key) {
                    // convert undefined to null because undefined data will be removed in convertDataToTargets()
                    var v = isUndefined(o[key]) ? null : o[key];
                    new_row.push(v);
                });
```

becomes

##### 0.4.11
```javascript
targetKeys.forEach(function (key) {
                    // convert undefined to null because undefined data will be removed in convertDataToTargets()
                    var v = $$.findValueInJson(o, key);
                    if (isUndefined(v)) {
                        v = null;
                    }
                });
```

I'm not exaclty sure what the **findValueInJson** does, but the part replacing brackets with dots and then spliting on the dots seems to be the cause of problems.

Anyway, until this issue is resolved, using the **0.4.10** versions solved it for me.

Cheers,
Kyhel

DoomyTheFroomy:
@Kyhel thank you very much! So I can use it, until the issue has been fixed!

kelthuzad:
This was implemented as a fix for issue https://github.com/c3js/c3/issues/1471 .
[...]

Issue #1471, opened 2015-11-19
Loading nested JSON objects


kelthuzad:
When I load JSON as chart data, I cannot load any nested objects or array values, but only objects in the first "level" of the JSON object.

This works fine:
```javascript
c3.generate({
data: [{myVal:1}],
keys: {value: ['myVal']} // load property "myVal" 
});
```
But this does not work:
```javascript
c3.generate({
data: [{myVal: {innerVal:1}}],
keys: {value: ['myVal.innerVal']} // cannot access nested data here
});
```
There would only be a small change needed to implement the access of nested values, and it would make loading of JSON objects much easier in some cases.

deeg:
+1

crearc:
+1

kelthuzad:
I just opened a pull request but the CI says my new test cases don't work. However they work fine on my machine. Any suggestions why they don't work here: https://travis-ci.org/masayuki0812/c3/builds/96603362 ?

tdharris:
I would like for this to work.

kelthuzad:
Just fixed my pull request, so tests work now!

z:
This looks great. Looking forward to this being merged.

aendrew:
This is a solid idea. Merging PR to `dev` now.

kelthuzad:
For anyone who is interested, this is a simple jsfiddle that demonstrates the new feature: http://jsfiddle.net/h0wag7ur/

michaelrodov:
Great feature! 
Will allow us to stop massaging the jsons prior to charts building

z:
Awesome! So happy to see this merged. Appreciate the demo too.

akshaysGit:
Hello,

I am trying to use following JSON data input for stacked bar chart, however its not working for nested input. If I put everything on one level, it works fine. Its just that nested is causing issue. can you pls help?

JSON data - 
`[
	{
		"Month": "Sample 1",
		"TotalAmount": "$100",
		"StartDate:
		"EndDate:		
		"Detail":
			[
				{ "Type": "A", "Unit": "50",  "Amount": "20"},
				{ "Type": "B", "Unit": "20",  "Amount": "70"},
				{ "Type": "C", "Unit": "2",  "Amount": "10"},
			]
	},
	{
		"Month": "Sample 2",
		"TotalAmount": "$150",
		"Detail":
			[
				{ "Type": "A", "Unit": "50",  "Amount": "20"},
[...]

PR #1694, opened 2016-05-12
Fix for issue http://github.com/c3js/c3/issues/1691

kelthuzad:
This is the bugfix for Fix for issue #1691 
The bug was introduced with the implementation of #1471

See http://jsfiddle.net/L3obmfsn/ for a working example.

If an object has key that contains . or [], it is now correctly found
instead of searching for an inner object or array that does not exist.
Example:
var data = [{"key.withdot":"1"}]; // did not work before, works now
var data2 = [{"key":{"innerkey":"1"}}]; // still works

Updated test with new bugfix.


tsileo:
Any update on this issue?

DoomyTheFroomy:
After I have run into the same problem again, could you please let this pass and release a new version? This would be a great new years present 👍 🥇

carmamir:
Would be great to include this bug fix into a main release as it is a big issue with any titles having dots (starting with domain names !!!)

scherrs:
Is there an update on this issue?

1Jesper1:
Can we merge this pr?

svki0001:
Anything new to merging this?

miladtai:
When will this problem be solved ?

kt3k:
closing in favor of https://github.com/c3js/c3/pull/2761

PR #1700, opened 2016-05-18
Fix: Failed to parse x '1463498940000' to Date object

brownman:
when parameter:date is an Array with 1 element of type number


panthony:
`parseDate` expect a single value to be coerced into a date.

If an array is provided at one point the issue is coming from the caller.

PR #1704, opened 2016-05-20
point.focus.expand.r can now take a function

### LINKED ISSUE #1571: Adds function handling for point's size on focus and select
The options `point.focus.expand.r` and `point.select.r` should both be able to take a function with either the data or the current point size.

Today when you use a function for `point.r` to make the point's size vary according to some data the default ratio of `4` and `1.75`.

I'd be happy to provide a PR if I knew it would be merged eventually.

panthony:
It fixes #1571


kt3k:
@panthony 
Thanks for the contribution!

Issue #1280, opened 2015-07-21
Change culling after zooming


revati:
Hello, i have question, is it possible to change **axis.x.tick.culling** value after user changes zoom by selecting some region in a sub chart?

- It probably shouldn't be to hard access **axis.x.tick.culling** value, but will it take new value to rerender x axis ticks?
- How can i get zoomed range that is used (how many values are shown). By browsing source it seems that it is called **domain**.

nitin0708:
I have also stuck from this.

catmeme:
+1 Would love a solution to this.

anudeepshetty17:
I also waiting for the solution for this can somebody tell me if this got resolved lately?

michaelrodov:
Hi there! 
We try to keep the issue queue clear for specific code-related problems with C3, 
Could you please try asking your question on either the [Google Group](http://groups.google.com/forum/#!forum/c3js) 
or [Gitter](https://gitter.im/masayuki0812/c3) instead? 

Thank you!​

catmeme:
@michaelrodov I understand the desire to clear out the issue queue but this is borderline a feature request.

I interpreted the issue to be requesting the c3 options or API support configuring tick culling values depending on zoom.

PR #1730, opened 2016-06-20
Culling to consider zoomed-in range rather than the whole range - implementing #1280

mihailik:
**#1280  Change culling after zooming** is needed for my use case, so here I am implementing it. Hope it helps!

Update: I've committed a better version, calculating `cullSetLength` instead of generating actual `cullSet` — saving array memory allocation.

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -546,8 +546,16 @@
         // show/hide if manual culling needed
         if ((withUpdateXDomain || withUpdateXAxis) && targetsToShow.length) {
             if (config.axis_x_tick_culling && tickValues) {
-                for (i = 1; i < tickValues.length; i++) {
-                    if (tickValues.length / i < config.axis_x_tick_culling_max) {
+                var cullSet = tickValues;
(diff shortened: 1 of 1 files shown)


mihailik:
@michaelrodov would you please consider this PR?

@revati @nitin0708 @catmeme @anudeepshetty17 hope that helps in your case, please have a look and give it a try

Many thanks!

c3.js
```diff
@@ -546,8 +546,20 @@
         // show/hide if manual culling needed
         if ((withUpdateXDomain || withUpdateXAxis) && targetsToShow.length) {
             if (config.axis_x_tick_culling && tickValues) {
-                for (i = 1; i < tickValues.length; i++) {
-                    if (tickValues.length / i < config.axis_x_tick_culling_max) {
+                var cullSetLength = tickValues.length;
```
ramdog: I believe you want to add this to `/src/zoom.js` not `c3.js` as this is the result of the build.

(same as https://github.com/c3js/c3/pull/1733/files#r69476315)

PR #1733, opened 2016-06-22
Invoke onzoom when zoom changed interactively

mihailik:
Currently [`onzoom`](http://c3js.org/reference.html#zoom-onzoom) is invoked when [`zoom()`](http://c3js.org/reference.html#api-zoom) function is invoked explicitly, but not when scale is changed due to mouse interaction.

Perhaps it's a bug, not intentional feature?

So, here's the fix for that!

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -5747,6 +5747,7 @@
         var $$ = this, z = $$.config.zoom_enabled ? $$.zoom : function () {};
         $$.main.select('.' + CLASS.zoomRect).call(z).on("dblclick.zoom", null);
         $$.main.selectAll('.' + CLASS.eventRect).call(z).on("dblclick.zoom", null);
+        $$.config.zoom_onzoom.call($$.api, $$.x.orgDomain());
     };
     c3_chart_internal_fn.redrawForZoom = function () {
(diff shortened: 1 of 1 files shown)


c3.js
```diff
@@ -5747,6 +5747,7 @@
         var $$ = this, z = $$.config.zoom_enabled ? $$.zoom : function () {};
         $$.main.select('.' + CLASS.zoomRect).call(z).on("dblclick.zoom", null);
         $$.main.selectAll('.' + CLASS.eventRect).call(z).on("dblclick.zoom", null);
+        $$.config.zoom_onzoom.call($$.api, $$.x.orgDomain());
```
ramdog: I believe you want to add this to `/src/zoom.js` not `c3.js` as this is the result of the build.

mihailik: Will do, thanks for the tip!

missingdays:
Hey! You still didn't move this change to src/zoom.js. After you do that, maybe you can write simple test, that will demonstrate this problem, and it will show that you have fixed that? If you need some help with testing, I can help you with that!
After that, I think this PR should be merged in.

kt3k:
now zoom.onzoom is being called each update invoked by scroll interaction. cf: https://github.com/c3js/c3/blob/v0.6.7/src/zoom.js#L41

PR #1741, opened 2016-06-28
Avoid random sort of legend

EijiSugiura:
Accidentally the order of legend are changed, when we create a graph which both of tooltip and legend are enabled.
To avoid this behavior, we pass deep copy of "$$.data.targets" to orderTargets().

Though it is dirty fix, but works for us.


craigdrayton:
I'm experiencing this problem too: I can load data and then order it using an order function, but as soon as I resize the chart, the legend order changes.

dborstelmann:
I am also experiencing this problem.  Tested out this fix and it works for me.  Would love to see this get merged soon so I don't have to use my own copy of c3

PR #1753, opened 2016-07-05
Add chart.width() and chart.height() API. Closes #1752

### LINKED ISSUE #1752: chart.width() and chart.height() API
As stated in #1644 there is no API for changing only width or height of chart at the moment.

missingdays:
Add new chart.width() and chart.height() API methods for making resize easier.
This closes #1752


PashaShulga:
How to use for resize chart?

missingdays:
@PashaShulga width() and height() are both setters and getters
So chart.width(someNumber) will resize chart to someNumber width. The same goes for height.
If you need to get current width, you just call chart.width().

SlyTheKing:
This would be extremely helpful :+1:

panthony:
@missingdays I'd argue that maybe a size() getter/setter that takes `{ height: X, width: X }` would be better ?

For the a lib that is already struggling with performance reason, providing an API that allows updating width and height at the same time and do a single flush may be better.

What do you think?

missingdays:
@panthony there is already API for resizing using both width and height
https://github.com/c3js/c3/blob/master/src/api.chart.js#L3

Separate width and height API makes it easier to change one property without getting the other. So instead of
```
chart.resize({'width': 500, 'height': chart.config.size.height})
```
you could just write 
```
chart.width(500)
```

panthony:
@missingdays `resize()` should clearly handle `resize({width: 100})` or `resize({height: 50})`.

And in this case `width(100)` could simply be an alias of `resize({ width: 100 })`.

missingdays:
@panthony I might be wrong here, but if you don't pass the other property in `resize`, it is set to null, which means "auto-size", as `getCurrentWidth` [here](https://github.com/c3js/c3/blob/master/src/size.js#L5) calls `$$.getParrentWidth`

panthony:
@missingdays You're right, what I had in mind would change the current behavior which is not good.

Well, if you could just update the documentation then to add `height()` and `width()` to the documented API.

The documentation site is now part of this repository.

missingdays:
Added docs for width and height to reference.html.haml

Issue #1763, opened 2016-07-13
C3 Line chart - Y axis issues


lorillcrees:
Hi, I am having problems with the Y-axis functionality. I have a line chart, and I wish to have the following:
- a min value of zero
- Y axis with 3 ticks
- first tick = 0
- padding between the first tick (zero) and the base line

I can't seem to get all of this desired functionality into one chart. Please see my JSFiddle:
https://jsfiddle.net/lcrees/55ea7k3j/

As you can see in the fiddle, I can have a min value of zero with the first tick = 0 and padding between the first tick (zero) and the baseline, but only if I do not set the number of ticks I want (see Chart 1). Once I set the number of ticks, the default padding is lost and the ticks do not start at the min value of zero. (see Chart 2)

Also, from our dynamic charts in our application, I notice that some random data points cause the Y axis first tick to be set to a negative number despite the min being set to zero (see Chart 3). 

Is there a way to achieve all our desired functionality? Basically, the functionality we want would be Chart #1 but with only 3 ticks.

Thanks

PR #1789, opened 2016-07-31
Adding support for y-axis tick count following d3 algorithm

mayrop:
Issue related: https://github.com/c3js/c3/issues/1763

Tested:
https://jsfiddle.net/zqrs2jt7/1/

And here's an example from data from the karma tests:
https://jsfiddle.net/y8tmgv6n/3/

By adding `count: N`, `exact: false`, will force to use this function and call d3 algorithm: 
https://github.com/c3js/c3/blob/master/c3.js#L7038

which calls `d3_scale_linearTicks` function using `d3_scale_linearTickRange` algorithm:
https://github.com/d3/d3/blob/v3.5.17/d3.js#L7582

It won't be always the exact number specified in the `count`. Per documentation:

https://github.com/d3/d3-3.x-api-reference/blob/master/Time-Scales.md
If _count_ is a number, then approximately _count_ ticks will be returned. If _count_ is not specified, it defaults to 10. The specified _count_ is only a hint; the scale may return more or fewer values depending on the input domain.

### DIFF
--- a/spec/axis-spec.js
+++ b/spec/axis-spec.js
@@ -65,6 +65,36 @@ describe('c3 chart axis', function () {
 
     });
 
+
+    describe('axis.y.tick.exact', function () {
+
+        it('should update args to have 12 ticks on y axis', function () {
+            args.axis.y.tick.count = 12;
+            expect(true).toBeTruthy();
+        });
+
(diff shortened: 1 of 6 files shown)


panthony:
May somewhat address #2210 also

Issue #1764, opened 2016-07-13
The ordering of the data points in tooltip is always in descending order


paulsoft:
version: 0.4.11
browser: any

The ordering of the data elements are in terms of their value. 
Eg: if there are.. 
data1: [10, 20, 30]
data2: [15, 5, 32]

then in the 1st point the order in tooltip is data2, data1. In 2nd data1, data2. in 3rd data2, data1.

But it will be harder for the person who is analysing graphs to see the data1 in different position in the tooltip.

As far as I've checked there apis of yours, I didn't get one to fix the order of data in the tooltip.

This thing should be implemented I think

TJWilson92:
It looks like this issue was fixed in pull #1813 - can we close it ?

Issue #1050, opened 2015-03-08
Tooltip options


netAction:
An option to group tooltip values only when they are in a small range would be nice. This is important since I don't have the cursor's value in _tooltips.contents_ option to fix it myself.

It should be possible to order the entries of grouped tooltips by value. Is there a way to run _getTooltipContent()_ from _tooltip.js_ in a _tooltips.contents_ option to achieve this behaviour manually? If not, it would be great to have a _tooltips.contents_ example with the default behaviour.

markdboyd:
I'm not sure if this is what you are looking for, but there is the ability to override the tooltip contents function via an option:

http://c3js.org/reference.html#tooltip-contents

In terms of the default behavior for tooltips, I think this covers it: 

https://github.com/masayuki0812/c3/issues/444#issuecomment-53986614

kriticism:
![image](https://cloud.githubusercontent.com/assets/10485230/8472656/b6283592-20c0-11e5-845b-9f9942ef47c0.png)


It seems that using 'data.order : null' has caused the shown stacking bottom-to-top in the stacked chart, top-to-bottom in tool-tip, and left-to-right in the legend.
I want to change the order of contents displayed in the tool-tip. Like VNe on top, then Ne, then N and so on. Or any other customization (which, say, is stored in an array). 
How do i do that?

michaelrodov:
Hi there
If this Is still relevant please provide a more detailed description of the desired functionality, including drawings when needed and use cases.
thank you for understanding

aashish-chaubey:
Hey people I am having an issue with the tooltip. I did not open a separate issue for this because it seems to be an responsiveness issue. 

![tooltip-issue](https://user-images.githubusercontent.com/12701463/39909808-fce0144a-5511-11e8-8d38-d801513497ad.png)

Please look into the above issue.
The tooltip text goes on the next line. How to fix this.

Thanks

Sibasish19:
Hey people, I am trying to add a functionality in my line chart create using C3. I want to hide the tooltip when the value is 0 and show it only for data points having non-zero values, is it possible? Please suggest me how to accomplish this. Thanks.

aaronELBorg:
I too would like the functionality mentioned by sibasishm.  Has it been added?  Thank you.

PR #1813, opened 2016-08-19
Fix tooltip ordering with data_order and add tooltip_order

panthony:
This is an attempt to fix various issues around tooltip ordering ( #1050 #1764 ).

This PR adds:
- handling of `data_order` with array (follow the order given by the array)
- new `tooltip_order` option that takes the same arguments as `data_order` (except for functions where parameters are values instead of data)
- tests that hopefully cover the majority of cases

This PR fixes:
- sort of tooltip if `data_order` is `null` or a `function`

### DIFF
--- a/spec/tooltip-spec.js
+++ b/spec/tooltip-spec.js
@@ -2,16 +2,20 @@ describe('c3 chart tooltip', function () {
     'use strict';
 
     var chart;
-    var tooltipConfiguration;
+    var tooltipConfiguration = {};
+    var dataOrder = 'desc';
+    var dataGroups;
 
     var args = function () {
         return {
             data: {
                 columns: [
-                    ['data1', 30, 200, 100, 400, 150, 250],
-                    ['data2', 50, 20, 10, 40, 15, 25],
-                    ['data3', 150, 120, 110, 140, 115, 125]
+                    ['data1', 30, 200, 100, 400, 150, 250], // 1130
+                    ['data2', 50, 20, 10, 40, 15, 25],      // 160
+                    ['data3', 150, 120, 110, 140, 115, 125] // 760
                 ],
+                order: dataOrder,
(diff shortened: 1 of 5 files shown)


pchaganti:
+1

bchartoff:
Would be great functionality, if a merge makes sense to y'all!

phandel:
+1

tcunha85:
+1

jorgemmsilva:
+1

aendrew:
Apologies for the delay in getting to this — merging!

alexgerv:
@aendrew this was merged onto dev. Should we open a new PR for master ?

aendrew:
@alexgerv Yeah, if you wouldn't mind that would be terrific — apologies for the hassle!

alexgerv:
No worries @aendrew , see PR #2070

chaitanmk:
Hi,

can you please provide sample code for tooltip_order.

CaffeineFusion:
@chaitanmk - had a flick through the source and derived the following:

// Order by value
tooltip: {
	order: 'asc'
}

//Order by label
tooltip: {
    order: function (t1, t2) {
        return t1.id > t2.id;
    }
}

**Note:** The sorting for tooltips appears to work in reverse of the order functions for data labels.
data: {
    order: function (t1, t2) {
        return t1.id < t2.id;
    }
}

Corresponds to the opposite function for the tooltip sort:
tooltip: {
    order: function (t1, t2) {
        return t1.id > t2.id;
    }
}

easz:
using c3@0.7

so basically if I want the data shown in tooltip having the same order as that in data `columns`, I have to extra declare this:
~~~
tooltip: {
    order: function(t1, t2) {
      return t1.index > t2.index;
    }
~~~

why we cannot just make it as default?

PR #1815, opened 2016-08-23
Added "bar.space" parameter to set space between bars in bar charts

panthony:
This PR is heavily inspired by #1328.

The difference is that the bar's space is subtracted from the bar's width.

That way the bars container will not be bigger than the container with no space between them.

It also adds some UTs (could be more  exhaustive with fixed bar width, bar ratio, though...)


kt3k:
Thanks for the contribution!

This has been merged through #2146. Released at v0.4.17.

Thanks!

PR #1818, opened 2016-08-29
Call onhide on each tooltip hide

johnsudaar:
The onhide method of the tooltip api is only called when the tooltip is closed from the api.
This PR call this methods when the tooltip is closed from the api or from an internal reason.


impronunciable:
@johnsudaar thank you for your contribution.

Before merging this we'll need you to add automated tests for this issue.

ShaneXie:
and same thing to onshow event, fix please. :)

Issue #1843, opened 2016-09-19
😱 Over a third of unit tests are rubbish!


aendrew:
I think part of why we've had so many regressive bugs added is because we're not actually unit testing huge swaths of the codebase.

I ran into the following line while troubleshooting an issue for a user:

`expect(true).toBeTruthy()`

This assertion *will always pass*. And it's *littered* throughout the codebase.

https://github.com/c3js/c3/search?utf8=%E2%9C%93&q=%22expect%28true%29.toBeTruthy%28%29%22&type=Code

If you run `grep "expect(true).toBeTruthy()" spec/* | wc -l`, you get `140`. That means of the ~400 assertions made across all unit tests, 140 *mean absolutely nothing*.

I repeat:

# 140 of our unit tests test _absolutely nothing_

This is just one invalid assertion. For all we know, even more unit tests don't actually test what they're intended to test but look like they do. 

Part of this is how C3 was originally constructed — because of a lack of modularity and how interconnected each piece is (particularly due to a reliance on concatenation), it wasn't really that easy to import a particular part of the library and run unit tests against it — you had to deal with the entire processed library and any additional side effects caused by that.

I'm not sure what the best answer is here. My thought it to just bin any unit tests with `expect(true).toBeTruthy` in them so as to reduce the coverage score, convert everything to Mocha as per #1838, and then add the appropriate tests, gradually increasing test coverage.

aendrew:
Update: doing a global replace on the spec/ folder with `expect(false).toBeTruthy()` takes us from 6 failures on `master` to *145.*

I'm honestly considering committing these improved assertions to `master` so we at least have a better idea of what unit tests we need to fill out...

pchaganti:
👍

xaranke:
🇬🇧

gouegd:
Another possibility is to change all those useless tests be inside a `it.skip` (or `describe.skip`) instead of the current standard function with no "skip".
This way the tests will appear as having been skipped, rather than incorrectly being reported as successful or failing (it should really be neither).
I'm doing this with mocha, not sure what you're using but it's probably available for you as well.

aendrew:
**Update:** I'm going to use this issue to coordinate efforts to fix unit tests and increase test coverage. @gouegd's suggestion to change misleading tests to skip is a good one and I'm going to start with that.

aendrew:
Invalid tests now skip. 

[...]

PR #1844, opened 2016-09-19
Changes all stubbed assertions to fail

aendrew:
See #1843. Because these cases have more than just the assertion, they won't fail if these assertions are simply removed. As such, I've explicitly set them to fail so we have a better idea of what our test coverage is. It's going to be _well fun_ getting back to a state where tests pass... :weary:

This shouldn't probably be merged for the moment until we agree this is the best course of action at the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ before_script:
 script:
 - grunt
 
-after_success:
-- npm run codecov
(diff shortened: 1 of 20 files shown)


kt3k:
Closing in favor of #2050

PR #1855, opened 2016-09-30
Fix #1506 Grid not aligned with ticks

### LINKED ISSUE #1506: Grid not aligned with ticks
Grid lines don't seem to be aligned with the ticks even in the non zoomed view.

gridlines non aligned

mistic100:
Add new `grid.x.ticks`  options :
- integer value (default: 10) -> approximate number of lines to display
- 'auto' -> display lines aligned with axis ticks
- array -> manually define the lines

### DIFF
--- a/htdocs/index.html
+++ b/htdocs/index.html
@@ -233,15 +233,21 @@
                 Show label on data with format
               </a>
             </div>
-            <div class="col-md-4">
-              <h3>Data Region</h3>
-              <a href="./samples/data_region.html">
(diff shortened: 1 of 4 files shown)


Edurne1:
Hi!

Is there any example of ths property 'grid.x.ticks'?

panthony:
@mistic100 Would you be kind enough to provide a test with your fix?

That would be super awesome.

I know... It's 2018.

Issue #1044, opened 2015-03-05
Multiline x-axis labels interfere with the bottom legend + proposed solution


smuuf:
I'd like to propose a fix that would handle the mess that emerges when one is having multiline labels on the x-axis while having the legend positioned at the bottom. *(which happens at least with the newest Chrome)*

**The problem:** *(don't mind the labels repeating - it's just to make them longer)*

![image](https://cloud.githubusercontent.com/assets/6860713/6506007/bb357648-c347-11e4-9e7c-81466bce1b65.png)

After a few hours of figuring out how to do this I came up with a rather simplified and ***naive solution***, that consists of 4 actual changes to the c3.js file:

Step 1: **c3.js, lines 128 - 132**
```
$$.currentMaxTickWidths = {
	x: 0,
	y: 0,
	y2: 0
};
```
...replaced with...
```
$$.currentMaxTickWidths = $$.currentMaxTickHeights = {
	x: 0,
	y: 0,
	y2: 0
};
```

===

Step 2: **c3.js, line 345**
```
xAxisHeight = config.axis_rotated || hasArc ? 0 : $$.getHorizontalAxisHeight('x'),
```
...replaced with...
```
xAxisHeight = config.axis_rotated || hasArc ? 0 : $$.getHorizontalAxisHeight('x') + $$.getMaxTickHeight('x', true),
```

===

Step 3: **c3.js, insert new method after the body of ```c3_chart_internal_fn.getMaxTickWidth``` method**
```
c3_chart_internal_fn.getMaxTickHeight = function (id, withoutRecompute) {
	var $$ = this, config = $$.config,
		maxHeight = 0, targetsToShow, scale, axis, body, svg;
	if (withoutRecompute && $$.currentMaxTickHeights[id]) {
		return $$.currentMaxTickHeights[id];
	}
	if ($$.svg) {
		targetsToShow = $$.filterTargetsToShow($$.data.targets);
		if (id === 'y') {
			scale = $$.y.copy().domain($$.getYDomain(targetsToShow, 'y'));
			axis = $$.getYAxis(scale, $$.yOrient, config.axis_y_tick_format, $$.yAxisTickValues);
		} else if (id === 'y2') {
			scale = $$.y2.copy().domain($$.getYDomain(targetsToShow, 'y2'));
			axis = $$.getYAxis(scale, $$.y2Orient, config.axis_y2_tick_format, $$.y2AxisTickValues);
		} else {
[...]

PR #1858, opened 2016-10-10
fix overlap between multiline category xAxis and legend or axis label

panthony:
This is an attempt to fix part of the issue around x axis tick values overlapping the legend or the x axis label.

I say "part of the issue" because I only tried to fix the case where the X Axis is of type "category" (since the ticks are displayed/aligned differently).

Thanks to #1044 that helped me get started.

Would have write some tests but I'm lacking time right now and does not look like PR are getting handled.


panthony:
(Failing tests were failing before my PR)

Issue #1648, opened 2016-04-11
D3 4.0 compatibility


gabrielrtakeda:
Is there compatibility with D3 4.0?

gabrielrtakeda:
I just installed C3 on version `0.4.11-rc4` and D3 on version `4.0.0-alpha.29` and seems it don't work. :(

abacaj:
No it does not work there are some breaking changes in D3 4.0 and as such will require that c3.js calls be updated as well.

aendrew:
Interesting. Perhaps this is a good opportunity to rewrite much of the codebase to be more modular, such as I'm researching on [my Webpack fork](https://github.com/aendrew/c3/tree/webpackify). Because everything is now using ES6 imports with Rollup in D3, it might be really smart to enable things like tree shaking with C3 through better use of modules.

That said, I'm nowhere near a JS design pattern pro — if you look at my fork, I've basically just imported each file as CommonJS, keeping to how the concatenation process builds the library. Any suggestions how to refactor to be more modular would be much appreciated.

abacaj:
I've been contemplating forking c3, I am using it heavily for internal tools and have actively maintained my own version of it. I've had to fix many bugs in it and unfortunately, I knew that any commits to this repo would not go unaccepted. 

Some of the things I've added are caching of attributes and selectors, this improved performance by two fold on the chart redraw methods and initial render. I've also had to add features like the capability to render gradient charts and better tooltips. Countless other undocumented fixes that I had to implement while using this library. 

It would make sense to have c3.core and then all charts can be included to produce a super lean build, essentially only having the user to require the core lib and the charts they need, in my case I'm only using 4 of the charts in c3.

Not to mention the D3-4.0 improvements that could be included in c3. Anyway I can rant on about this library but it is one of the best ones out there just sad we cant contribute to this one.

amenadiel:
I've played around with c3 and d3@v4. Some core d3 namespaces do no longer exist (d3.behavior, for example) so I couldn't get past that. 

However, I managet to get some Rollup sweetness by replacing c3's concat and d3 smash tasks to use ES6 imports and exports. That doesn't leave much room for three shaking, but I'm not done tinkering.

@abacaj  what I have done in the past was fork into an organization I made just to hold my open source projects and long term forks:  https://github.com/huaofoundrie. If you're interested tell me and I'll give you admin permissions so we could mantain c3 nextgen from there

aendrew:
[...]

PR #1871, opened 2016-10-31
Version 1.0 [work-in-progress]

aendrew:
This PR is meant to rally efforts towards getting to v1 of C3.js.

Key features of v1:
- [ ] D3 v4 compat (#1648)
- [ ] Charts provided via plugin system
- [ ] Component parts separated into ES2015 modules
- [ ] Each part of internal API fully documented
- [ ] TypeScript definition generated from documentation


panthony:
Hi @aendrew,

Is this project still being worked on ? 

I really like this library I'm using it a lot and I'd rather invest time fixing and adding missing things that redoing everything using another chart library.

But looking at the issues that are never closed or responded to, PR that are never merged, and this big bang PR that will break every other ones I wonder if it is still worth it to try to fix things.

Maybe I could help in some way ?

aendrew:
@panthony I'm honestly not sure what the best approach is at this point. I've admittedly halted work on this PR for the moment as I'm drowning in work and behind deadline on my book, but am hoping to return to it at the start of next year with fresh eyes. I think what's needed more than anything is a feature freeze until we can get everything sorted out architecturally, so if you're wanting features C3 lacks, you may want to look elsewhere for the moment, unfortunately.

panthony:
@aendrew Well I could always workaround the issues, fix them in a fork, or directly implement what I need using D3 directly.

What worries me is more the issues not being closed, PR proposing unit tested fix not being merged, etc.

I'd like to avoid diverging too much from the official repository.

And are you alone on c3 ?

When I walk around issues because I'm looking for something I often see duplicate issues, or simply questions about "how do things" that could be closed since they should use the google groups instead.

And I honestly think that the scope of this PR is way to broad and should be broken down into several steps in order to not freeze everything for months. Isn't it possible to move code as module little by little ?

aendrew:
@panthony Good points. I've not gotten a lot of input or buy-in from the other contributors so I don't perceive there to be a lot of appetite for my current tack with this. My feeling is that it might be better to close this PR for now to prevent confusion, re-evaluating efforts to get to v1.0 in the New Year.

PR #1877, opened 2016-11-03
Fix to consider existing axis X custom values when the custom id is not at the index 0. fixes #1876

### LINKED ISSUE #1876: x axis custom values being ignored and using the value index instead.
Hello.

I have a situation where I specify two custom x axis values array, and when building the second array, it starts to use the x value index instead of the value.

I've debugged the problem and found it to be located on the data.convert.js file (c3/src/data.convert.js#L179), as it doesn't check if the X value already exists before getting its id.

Thanks in advance.

mfklauberg:


leyendecker:
+1

herberttn:
+1

Issue #1357, opened 2015-08-29
data.regions bug when zoom: enabled


wansinn:
Hey,

first of all thanks for the awesome work with this lib!

I ran into an issue / bug? when using regions with enabled zoom:
The line with regions in it doesn't change when zooming, the points do.
You can see the issue here: http://jsfiddle.net/xqjb2zhb/
- One line has regions and doesnt zoom
- the other line has no regions and zooms properly.

Another question:
Is it possible to define "connectNull" on a per line base?
For example: I have several lines with null vales. Some I would like to be connected inbetween, some not. (if I should ask this as a seperate question just throw me a hint and I will do so).

cheers,
oli

jrsala:
Same issue here, any line with regions specified on it does not react to zooms.

Thank you for your work!
J.

sidiee:
Do we have any update on this issue ????

jrsala:
I needed a region style to make the end of a line dashed (to represent the uncertainty of the future) and instead I has to use a region colored overlay, which is not as practical and appropriate as dashed lines but still OK.

I guess I'll just do all this in D3 myself when I have time?

gearoidc:
Same issue here. Attached Gif to show issue. 

![screen capture on 2016-03-04 at 14-11-01](https://cloud.githubusercontent.com/assets/65091/13529493/7054fa44-e214-11e5-803f-7d4cb107256e.gif)

chetanmotamarri:
Still facing the issue. Any update with this ?

panthony:
Fix is merged but no ETA for the next release.

Mshardul:
It's been almost 2 years, since the fix. Any updates on the next release date?

baconinthemorn:
I can confirm I am also still witnessing this issue where regions are not updating/reacting to when a chart is zoomed (with v0.7.15). Any ideas for a workaround in the meantime?

Lionking93:
I also encountered this issue with version 0.7.15. From inspecting the region elements I noticed that the g elements with the c3-region class do resize but the rect elements inside the g elements for some reason do not. I managed to create a workaround by specifying a callback function to the onzoom property (https://c3js.org/reference.html#zoom-onzoom) and modifying the rect element width and x attributes with d3 to match the width and x attributes of their parent elements.

    config.zoom.onzoom = () => {  
        let regions = d3.selectAll("#chartId .c3-region")._groups[0];  
        regions.forEach(regionElement => {
            let region = d3.select(regionElement);
[...]

Issue #1061, opened 2015-03-13
Dashed styling breaks when zoom enabled


mattlucas121:
When combining the zoom functionality with the "dashed" styling of a region the chart gets the styled X axis stuck on it and never redraws. Here is a jsfiddle showing the issue..

http://jsfiddle.net/ZpBdL/5/

This maybe related/similar to issue #1036. If you need any additional feedback please let me know.

Thanks!

tclavijod:
+1 .. I'm having the same problem :S

lejard-h:
:+1:

petogriac:
same issue, is there any fix ? 

thanks

panthony:
Fix is merged but no ETA for the next release.

kt3k:
The fix is released in [v0.5.2](https://github.com/c3js/c3/releases/tag/v0.5.2).

PR #1891, opened 2016-11-22
Use proper scale, fixes scaling issues with data regions  

galindus:
#1061
#1357


javflores:
Tried in the context of my chart and it works perfectly. 
Hopefully we can get your fix merged soon.
Cheers for looking into this!

PR #1914, opened 2016-12-24
Fix for issue #1873

### LINKED ISSUE #1873: Failing to parse moment dates
#@Tryton web client Sao use C3 to render graph views.

With version 0.4.10 timeseries works fine.

When we update C3 version from 0.4.10 to 0.4.11,
at display graph type line with timeseries, browser
show this error:
c3.js:1048 Failed to parse x '1478062800000' to Date object

We use momentjs date object to populate data.

0_4_10_version
0_4_11_version
cut

conceptualspace:
Attempt parsing object (ex Moment.js date object) to date. This was implicitly supported <= v0.4.10 but regressed with narrower validation in #1121.

---
close #1873

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1036,6 +1036,8 @@ c3_chart_internal_fn.parseDate = function (date) {
         parsedDate = date;
     } else if (typeof date === 'string') {
         parsedDate = $$.dataTimeFormat($$.config.data_xFormat).parse(date);
+    } else if (typeof date === 'object') {
+        parsedDate = new Date(+date);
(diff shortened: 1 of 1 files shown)


conceptualspace:
#1873

kt3k:
Seems reasonable to me. Thanks for the contribution!

If we have a problem about unexpected objects (other than moment), then let's use `._isAMomentObject` for detecting moment objects. https://github.com/moment/moment/blob/develop/src/lib/moment/constructor.js#L76

PR #1305, opened 2015-08-04
Fixes Uncaught TypeError: Cannot read property 'gauge_min' of null #1205

diesburgm:
Fix for issue #1205.

I added a console.log while debugging this to verify that this null condition was being hit and that the charts still generated correctly after hitting this. Duplicating this is very timing dependent.

Issue #1205, opened 2015-06-10
chart.destroy throws Uncaught TypeError: Cannot read property 'gauge_min' of null


Victori0us:
after redirect to another page and destroy view (when destroy view it is invoking and chart destroy) raise error: Uncaught TypeError: Cannot read property 'gauge_min' of null

benharden:
I'm having the same issue. Looking at the call stack, the top function is d3_timer_step, so I think the problem must be occurring when the chart is destroyed whilst a transition is still running.

grassick:
It also seems to permanently disable chart creation from working thereafter! Seems to happen with pie charts for me. At least I can't make a pie chart that has visible segments after it happens.

aendrew:
Seems at least three people have encountered this, meaning it's a pretty well-confirmed bug at this point.

diesburgm:
If it helps here is the full stack trace. If you destroy a chart while it is rendering the destroy sets all of the members to null, which then get accessed in the attrTween code being called on a timer from d3 to do the rendering.

Uncaught TypeError: Cannot read property 'gauge_min' of null
	c3_chart_internal_fn.updateAngle
	(anonymous function)
	attrTween
	(anonymous function)
	d3_class.forEach
	start
	(anonymous function)
	d3_timer_mark
	d3_timer_step

louy:
+1 @masayuki0812

aendrew:
#1205 has been merged, so am finally closing this issue! :D Many thinks to @diesburgm for the PR!

joshthecoder:
I seem to be running into this issue again but the stack trace is different.

<img width="634" alt="screen shot 2016-05-02 at 2 09 30 pm" src="https://cloud.githubusercontent.com/assets/25644/14968044/9a09f072-106f-11e6-804c-ae905beb8d1e.png">

Seems to happen on destroy while transitioning between views. The chart in the view
being presented next never loads.

aendrew:
@joshthecoder Mind providing which version of C3 you're using? If 0.4.11 (latest), could you please create a new issue and reference this one in it? Thanks!

sushantdhiman:
I am facing this same issue

```js
c3.js?84d1:2148 Uncaught TypeError: Cannot read property 'keys' of null
    at ChartInternal.c3_chart_internal_fn.convertDataToTargets (eval at <anonymous> (http://localhost:3000/build/bundle.js:7791:2), <anonymous>:2148:24)
[...]

PR #1949, opened 2017-01-31
Fixes Cannot read property 'keys' of null

pajcho:
When chart is destroyed while transitions are in progress we are getting Cannot read property 'keys' of null like many people reported as a part of #1205 

This solves issue in a same way like #1205 and #1305 are solved here https://github.com/FernCreek/c3/commit/a26b38165f4c0b97bb2f365978f3df9e12446530

### DIFF
--- a/c3.js
+++ b/c3.js
@@ -2144,113 +2144,115 @@
         return new_rows;
     };
     c3_chart_internal_fn.convertDataToTargets = function (data, appendXs) {
-        var $$ = this, config = $$.config,
-            ids = $$.d3.keys(data[0]).filter($$.isNotX, $$),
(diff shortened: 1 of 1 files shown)


ilacyero:
I think that will be better if you use:
```
    if (!config || !$$.d3) {
        return null;
    }
```
The change is really small and you will fix it too.

mervinva:
+1

panthony:
Hi,

Please modify the proper source file in `src` directory and do not commit the `c3.js` file which is overwritten by the build phase.

Also, Do what @ilacyero suggested and do an early return if config is not available anymore.

This is the same thing except it's more readable and introduce less changes.

Thanks

panthony:
No response & `Object.keys` is used instead of `$$.d3.keys`, closing

PR #1992, opened 2017-03-15
Style empty points in line charts via CSS

gcmznt:
moved point color from fill to color to let the possibility to use the color for stroke via css like: 

```
  .c3-circle {
    fill: white;
    stroke: currentColor;
    stroke-width: 2px;
  }
```

screen shot 2017-03-15 at 4 19 14 pm

### DIFF
--- a/src/scss/point.scss
+++ b/src/scss/point.scss
@@ -1,3 +1,6 @@
+.c3-circle {
+  fill: currentColor;
+}
 .c3-circle._expanded_ {
   stroke-width: 1px;
   stroke: white;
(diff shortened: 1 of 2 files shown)


stephencookdev:
+1, would really like to see this in the codebase

For anyone else landing here from Googling around, I'm currently doing the following hack to achieve a similar result to what this PR would provide:

```js
import c3 from "c3";

const originalRedrawCircle = c3.chart.internal.fn.redrawCircle;
const originalUpdateCircle = c3.chart.internal.fn.updateCircle;
c3.chart.internal.fn.redrawCircle = function() {
  const response = originalRedrawCircle.apply(this, arguments);
  const mainCircle = response[0];
  mainCircle.style("color", this.color);
  return response;
};
c3.chart.internal.fn.updateCircle = function() {
  const response = originalUpdateCircle.apply(this, arguments);
  this.mainCircle.style("color", this.color);
  return response;
};
```

panthony:
An up to date version exists here #2693

Issue #1862, opened 2016-10-19
Mouse position of onclick event


varuag:
Is there a way to get the mouse position of an onclick event on a data item? I'm trying to show a context menu when say a bar in a bar chart is clicked, and need the positon of the mouse to position the menu.

Thanks

panthony:
I believe you can simply do the following to retrieve the mouse's position:

```js
{
   data: {
    onclick: function(d) {
         const pos = d3.mouse(d3.event.target)
    }
  }
} 
```

The changes suggested by #2001 seems unnecessary (and wrong since I could not find any `event` within the scope of the function).

What do you think @kt3k ?

kt3k:
@panthony I agree with you. `d3.event` is the way to access the current event (ref: https://github.com/d3/d3-selection#event ), and #2001 seems wrong to me as well. Let's close it.

@varuag Please use the above example for using the mouse positions in click events.

PR #2001, opened 2017-03-29
Mouse position of onclick event

shlrur:
https://github.com/c3js/c3/issues/1862

user who use c3.js can use $event of mouse click event.


kt3k:
@shlrur Thanks for the contribution. Sorry for the late response.

Please see [this comment](https://github.com/c3js/c3/issues/1862#issuecomment-519063941) for accessing to the mouse position. We think this change is not necessary for getting the mouse positions.

PR #2030, opened 2017-05-09
Upgrade pathseg polyfill to 1.1.0

vitorbaptista:
The previous version used variables without declaring them first, which fails
because we're in strict mode. This change was introduced by https://github.com/progers/pathseg/commit/be4ff0dd7675f0221b726485f68af2565410ac99.

I found this issue while trying to import `c3` using `jsdom`.

### DIFF
--- a/src/polyfill.js
+++ b/src/polyfill.js
@@ -26,817 +26,849 @@ if (!Function.prototype.bind) {
   };
 }
 
-//SVGPathSeg API polyfill
-//https://github.com/progers/pathseg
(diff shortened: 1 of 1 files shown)


vitorbaptista:
FYI, I found another issue with `pathseg`, so this PR isn't enough to allow requiring `c3` inside the `jsdom` (see https://github.com/progers/pathseg/issues/19)

kt3k:
Thanks!

PR #2037, opened 2017-05-17
Switch to rollup

kt3k:
- This PR use rollup for bundling c3 instead of concat

In addition, this PR moves some grunt tasks to npm run scripts. I'm planning to move all task to there later PRs.


hjemmel:
WOW it is a great improvement, amazing work!

pacahon:
@kt3k It's really encouraging to see commits from one of the contributors. :) PR's tab has looked like a cemetery until now. Can we dream you plan to little bit revive support of this library? I'm in doubts, should I create PR's or issues in case when nobody have time to review them.

kt3k:
I'll check by hand all the examples in c3js.org are ok before merging this just in case.

@pacahon We're trying to restart the timely maintenance like before. Please see [commit history](https://github.com/c3js/c3/commits/master) for now. I believe test improvement is key for that.

.jshintrc
```diff
@@ -2,7 +2,7 @@
     "bitwise": true,
     "eqeqeq": true,
     "curly": true,
-    "strict": true,
+    "strict": false,
```
aendrew: What's the rationale for setting this to `false`?

kt3k: When I use rollup with `--format umd` option the output script starts like the following

```
(function (global, factory) {
	typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
	typeof define === 'function' && define.amd ? define(factory) :
	(global.c3 = factory());
}(this, (function () { 'use strict';
```

And this breaks the `strict` rule of jshint. I couldn't configure rollup to insert `'use strict'` in the first function, so I disable it for the moment.

kt3k:
@aendrew Thanks for your review :)

.jshintrc
aendrew: @kt3k That makes sense; also, I forgot using ES6 modules implicitly turns "use strict" on, see https://stackoverflow.com/a/31685340/467760

aendrew:
Cool, done. Let's handle individual bug fixes on a per-PR basis, updating them as necessary to the new infrastructure.

@kt3k In case merging breaks anything you're working on, please feel free to revert the merge and then re-merge once ready.

kt3k:
@aendrew No problem. Thanks for merging!

iocube:
Hi.
CONTRIBUTING.md is outdated. Tried to follow the instructions and I failed to build for the following reasons:

- `Gruntfile.coffee` no longer contains 'default' task so running `grunt` fails.
- `grunt build` task fails because there is no such task `concat` anymore

After finding this PR, I understand now I should use `npm build`.

aendrew:
@iocube Want to be totally awesome and PR that for us? 😄 

Thanks!

iocube:
I'll create a PR today after work :wink:

PR #2044, opened 2017-05-27
Fix prepublish script

kt3k:
Master's build is failing because of prepublish script's bug.

I set prepublish script `grunt` at #2041 but also I deleted grunt `default` task in #2037 . This PR resolves the conflict between those 2 changes.


kt3k:
This is a hotfix of master's build and ci is all passing. I'm merging this.

PR #2045, opened 2017-05-27
Add `serve` script

kt3k:
It's convenient to have `serve` script in package.json. This doesn't require for contributors to have python installed.

In addition, I dropped `grunt-contrib-concat` and `grunt-karma`. These became unused by #2037.


aendrew:
LGTM! Not requiring Python is a good shout. I've fixed the merge conflict caused by #2049, will merge now.

PR #2046, opened 2017-05-27
Sets all stubbed unit tests to 'skip'

aendrew:
This will make invalid tests emit "skip" instead of passing (or failing as I initially suggested in #1843).

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -16,18 +16,9 @@
     "bitwise": false,
 
     "browser": true,
+    "jasmine": true,
 
     "globals": {
-        "jasmine": false,
-        "describe": false,
-        "beforeEach": false,
-        "beforeAll": false,
-        "it": false,
-        "expect": false,
-        "d3": false,
-        "module": false,
-        "require": false,
-        "define": false
+        "d3": false
     }
-
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -8,7 +8,7 @@
     "undef": true,
     "unused": true,
     "indent": 4,
-
+    "esversion": 6,
     "eqnull": true,
     "expr": true,
     "newcap": false,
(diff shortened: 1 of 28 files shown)


aendrew:
I've added a few more commits to fix the linting errors that are breaking Travis, and in doing so, am now linting the files in `src/` instead of the built c3.js file (which doesn't really actually need to be linted; we only care about code style in the composite source files). 

In doing so, I changed:

https://github.com/c3js/c3/pull/2046/files#diff-7eb52b366866677666470e019283c8eaL11

to:

https://github.com/c3js/c3/pull/2046/files#diff-7eb52b366866677666470e019283c8eaR1

This may or may not be problematic; do we want to let users supply D3 separately, or via CDN? My feeling is not; if they're using Webpack or Rollup to bundle their code, importing it via ES6 imports will allow better tree shaking. @masayuki0812 or @kt3k — any thoughts?

aendrew:
Okay. I've had a realisation.

The `expect(true).toBeTruthy();` specs are *actually needed for the tests to pass*. This is because they set various chart arguments before the rest of the assertions are run. :weary:

This is really confusing and not how to write unit tests — it's really hard to reason about how tests should work and makes testing feel really nondeterministic. 

I almost feel we should redo practically all of the tests to specifically test what each module does, using proper stubbing techniques, and then maybe create a separate functional testing suite. 

Any thoughts?

aendrew:
I'm closing this because skipping the bad tests breaks all the gosh-durn tests.

kt3k:
I agree that `should update args` things are not test cases. I think those should go into `beforeAll` clauses (See https://jasmine.github.io/2.1/introduction#section-10 ) because they do the preparation for later test cases and `beforeAll` is a function exactly for that purpose.

PR #2048, opened 2017-05-28
Improves linting

aendrew:
Previously, linting was done on the generated file and not on the component modules. This makes jshint now lint the modules instead and fixes any outstanding linting errors.

Note, this also now simply imports D3 in src/core.js:

```diff
+import * as d3 from 'd3';
-var d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;
```
I think this is desirable but am open to discussion.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,3 +1,4 @@
+import * as d3 from 'd3';
 import Axis from './axis';
 import CLASS from './class';
 import { isValue, isFunction, isString, isUndefined, isDefined, ceil10, asHalfPixel, diffDomain, isEmpty, notEmpty, getOption, hasValue, sanitise, getPathBox } from './util';
@@ -7,8 +8,6 @@ export var c3 = { version: "0.4.11" };
 export var c3_chart_fn;
 export var c3_chart_internal_fn;
 export var c3_chart_internal_axis_fn;
-
-var d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;
 
 export function API(owner) {
     this.owner = owner;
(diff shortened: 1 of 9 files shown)


src/core.js
```diff
@@ -1,3 +1,4 @@
+import * as d3 from 'd3';
```
kt3k: With the current settings of rollup, this part doesn't look up the global d3 in browser environment (This only works in commonJS and AMD environment). I think it's safe to keep that line as is.

kt3k:
Other part seems nice! Thanks for the contribution!

src/core.js
aendrew: That's sort of the question I was asking; it ultimately results in D3 being provided by the bundle and not a global — as such, users don't need to include D3 separately, just C3. If you're wanting to do something clever with C3 (i.e., use it as a dependency in a larger project), you should probably be using a bundler anyway.

Another possibility would be to set D3 as an external dependency; see https://github.com/rollup/rollup/wiki/JavaScript-API#external

aendrew: I've added D3 as a global/external to the Rollup instantiation; AFAICT, that's the intended way of using external modules with Rollup if you don't want to include them within the bundle itself. See also: https://github.com/rollup/rollup/wiki/JavaScript-API#globals

kt3k: I agree with that change (`--global d3:d3`). Thanks for finding it!

aendrew:
No idea why it's impacting CodeCov. If you're cool with adding D3 as a global, feel free to merge after I merge #2045.

Issue #2033, opened 2017-05-14
Refactoring Plan


kt3k:
### TODO
- [ ] set issue template 🔝 

### Pending
- [ ] switch to eslint or standard
  - a little bit future (this causes all PRs conflict)
- [ ] switch to mocha / power-assert / testdouble.js #1838 

### DONE
- [x] introduce greenkeeper 🔝 
- [x] introduce babel 🔝  ( in testing #2063 )
  - keeping chart_fn and chart_internal_fn at first
- [x] switch to real browser testing 🔝
  - first chrome #2083 
- [x] switch to circle-ci
- [x] move c3js.org to gh-pages
- [x] move fixture setting to `before` clause, not in `it` clause
  - Done in #2050
- [x] Remove Grunt
  - See #2067.
- [x] switch from grunt to npm run script #2064 #2065
- [x] switch to rollup 🔝  #2037

masayuki0812:
@kt3k 
Please see https://github.com/c3js/c3/projects/1 . There is already a nice project.

@aendrew 
Hi, how are you doing? As you can see, @kt3k, who is my ex‐colleague and good frontend developer, is interested in working on c3. So, I added him to the member and he will start working on the topics above. The topics he mentioned are common with the issues included in v1.0.0 project, so I guess it is not a problem to proceed basically.
I think joining a new member is good for c3 basically, but if there is something to discuss, please let me know.

aendrew:
@masayuki0812 This is amazing, the work @kt3k has done already is excellent — glad to see some new life brought into the project!

@kt3k Welcome aboard! I've already reviewed #2037 — we should merge it in sooner rather than later, I'd like to get some of this v1 stuff moving forward as soon as possible. 

Questions for both:

a. How should we handle the ongoing issue with the tests? If it's all the same to you both, I have quite a bit of free time at the moment and am wanting to try and resolve all the testing-related issues as soon as possible.

b. How should we manage all the outstanding PRs? Some are bug fixes, some are features. It's quite likely that merging #2037 will cause conflicts with most of the outstanding PRs. My feeling is we should just leave any bug fixes open and possibly close any feature requests, in the hope that their authors with resolve the conflicts — that said, until we have adequate test coverage, it's uncertain whether any given PR will cause regressive bugs (even if fixes another).

aendrew:
Updated the list slightly to include the Grunt removal.

kt3k:
All done except pending ones (lint and test framework switching).

PR #2049, opened 2017-05-28
Improves speed of Travis builds by not requiring sudo

aendrew:
This updates grunt-sass to now use node-sass, thus no longer requiring the Sass gem. 

It also disables sudo on Travis so builds run much faster, by including grunt-cli as a dev dep.

This is meant as a quick improvement until we migrate to CircleCI as per #2033.


kt3k:
The change seems good to me.

Strangely travis doesn't pass at the 350th case (I retried several times and the result is always the same).

aendrew:
@kt3k Odd; it seems to be timing out on `shape.line-spec.js`, which is the the only one that uses `parseSvgPath`. I'll investigate further.

kt3k:
Thanks for updating the timeout! LGTM. Merging.

PR #2050, opened 2017-05-28
🎉 All non-asserting specs removed. Fixes #2047.

### LINKED ISSUE #2047: Replace all expect(true).toBeTruthy() blocks with beforeEach blocks
See https://github.com/c3js/c3/issues/1843#issuecomment-304484663.

aendrew:
That was a painful use of three hours. :joy:

This makes the tests much easier to reason about and provides a path forward for getting rid of things like `spec/c3-helpers.js`.

Many thanks to @kt3k for suggesting `beforeAll` instead of `beforeEach` for all the non-asserting specs blocks!

### DIFF
--- a/spec/api.zoom-spec.js
+++ b/spec/api.zoom-spec.js
@@ -9,7 +9,7 @@ describe('c3 api zoom', function () {
 
     describe('zoom', function () {
 
-        it('should update args', function () {
+        beforeAll(function () {
             args = {
                 data: {
                     columns: [
@@ -22,7 +22,6 @@ describe('c3 api zoom', function () {
                     enabled: true
                 }
             };
-            expect(true).toBeTruthy();
         });
 
         it('should be zoomed properly', function () {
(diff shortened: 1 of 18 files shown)


spec/api.zoom-spec.js
```diff
@@ -41,48 +40,48 @@ describe('c3 api zoom', function () {
...
-                    x: 'date',
-                    columns: [
-                        ['date', '2014-01-01', '2014-01-02', '2014-08-01', '2014-10-19'],
-                        ['data1', 30, 200, 100, 400]
-                    ]
-                },
-                axis: {
-                    x: {
-                        type: 'timeseries'
+        describe('should load timeseries data', function () {
```
kt3k: I think something like `describe('with timeseries data', ` is more appropriate.

spec/arc-spec.js
```diff
@@ -47,149 +46,152 @@ describe('c3 chart arc', function () {
...
-                    .select('g.c3-shapes.c3-shapes-data2.c3-arcs.c3-arcs-data2')
-                    .select('path.c3-shape.c3-shape.c3-arc.c3-arc-data2'),
-                data3: chartArc.select('.c3-chart-arc.c3-target.c3-target-data3')
-                    .select('g.c3-shapes.c3-shapes-data3.c3-arcs.c3-arcs-data3')
-                    .select('path.c3-shape.c3-shape.c3-arc.c3-arc-data3')
-            };
-            expect(arcs.data1.attr('d').indexOf('NaN')).toBe(-1);
-            expect(arcs.data2.attr('d').indexOf('NaN')).toBe(-1);
-            expect(arcs.data3.attr('d').indexOf('NaN')).toBe(-1);
+        describe('should set args with data id that can be converted to a color', function () {
```
kt3k: I guess `describe('with data id that can be converted to a color', ` is better.

spec/arc-spec.js
```diff
@@ -47,149 +46,152 @@ describe('c3 chart arc', function () {
...
-                        ['data', 8]
-                    ],
-                    type: 'gauge',
-                    fullCircle: true,
-                    startingAngle: Math.PI/2
-                }
-            };
-            expect(true).toBeTruthy();
-        });
+        describe('should update args to have a 180 degree gauge', function(){
```
kt3k: I prefer `describe('with a 180 degree gauge'`.

kt3k:
The entire structure of the change seems good to me. Thanks for the huge improvement!

I think some of the description titles are a bit strange because it starts with `should update ...`. I commented on some of them.

aendrew:
@kt3k I definitely sped through doing them at 2:00 in the morning so it's quite possible the `describe()` blocks need better descriptive text. 😄 

Will go through again looking for the text string "should update" and then update this PR.

PR #2052, opened 2017-05-28
Adds sourcemaps to testing; updates dependencies

### LINKED ISSUE #2051: Code coverage should evaluate module files, not built file
Given we no longer concatenate everything and have everything in lovely ES6 modules now, we should start running coverage reports on the modules and not the built code. This will enable better analysis of where to improve test coverage.

aendrew:
* Adds karma-sourcemap-loader
* Adds karma-preprocessor-rollup
* Moves testing Rollup building to Karma
* Enables sourcemaps in tests
* Fixes #2051

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -31,8 +31,7 @@ module.exports = function(config) {
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
     preprocessors: {
-      'src/**/*.js': ['rollup', 'sourcemap', 'coverage'],
-    //   'spec/**/*.js': ['rollup']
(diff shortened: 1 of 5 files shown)


aendrew:
This is passing now, but I don't think it's actually posting code coverage properly yet.

kt3k:
`c3.js` entry in `files` section should be removed because it's the bundled version and we don't want that file loaded if we build c3 at the start of karma.

I guess we need rollup-plugin-istanbul for instrumentation. Usually you can't use karma-coverage directly if you use bundlers.

aendrew:
@kt3k That is literally exactly what I was about to try next. :joy: 

Awesome, will do that and then update the PR. :+1:

Issue #2054, opened 2017-05-29
Conditionally show points


iocube:
Hello.

`point.show` is all or nothing. I can't control which points I want to show / hide.
What if it will be possible to pass a function which should return true / false, and based on that, we'll show or hide the point.

Basically, I am talking about modifying `core.js#c3_chart_internal_fn.opacityForCircle` to something like this:

```javascript
c3_chart_internal_fn.opacityForCircle = function (d) {
    var isPointShouldBeShown;
    if (this.config.point_show instanceof Function) {
        isPointShouldBeShown = this.config.point_show(d);
    } else {
        isPointShouldBeShown = this.config.point_show;
    }
    var opacity = isPointShouldBeShown ? 1 : 0;
    return isValue(d.value) ? (this.isScatterType(d) ? 0.5 : opacity) : 0;
};
```

So the usage example will be:
```javascript
points: {
    show: function(d) {
        if (d.index % 2 === 0) { return true; }
        return false;
    }
}
```

or
```
points: {
    show: true
}
```
note: The old behavior is preserved.

What you guys think?

kt3k:
Sounds like a reasonable enhancement to me. PRs are welcome.

@aendrew @masayuki0812 
How do you think?

iocube:
A co-worker suggested that I can achieve same behavior if I will use `point.r`

Turns out `point.r` actually can be a boolean **or** a function which expects a `datapoint` as an argument (by the way, it is not documented [here](http://c3js.org/reference.html))

So if this function returns 0, then the point will not be shown.

Is it still reasonable?

kt3k:
Thanks for pointing the above.

Personally I still feel it's reasonable to have function option for `point.show` because it feels like a hack to set radius=0 for hiding it.

However this is design issue, so I'm going to collect other maintainers' opinions somewhere later.

kt3k:
Implemented at https://github.com/c3js/c3/pull/2062

This will be released at v0.4.13.

PR #2062, opened 2017-06-02
Allow passing a function to point.show configuration (#2054)

iocube:
This will allow to display points based on a condition.


kt3k:
Thanks for the contribution!

kt3k:
Published in v0.4.13.

PR #2064, opened 2017-06-03
Use jshint-cli instead of grunt-jshint

kt3k:
This reduces the use of grunt and make it easy to manage module dependencies.


kt3k:
Thanks!

PR #2065, opened 2017-06-03
Use node-sass cli instead of grunt-sass

kt3k:
Similar to #2064, this reduces the use of grunt and use node-sass directly.

I removed `watch` task for the moment. I'll restore it later if anybody needs it.


aendrew:
LGTM :+1: — don't see any issue with merging so am doing so.

So now only linting/minifying is done through Grunt? That's awesome, it means we can probably similarly revert for those too and be able to remove Grunt entirely!

kt3k:
Thanks! :)

PR #2066, opened 2017-06-03
Restore the timing to store the d3 reference

### LINKED ISSUE #2060: parallel loading of d3 and c3 broken since rev ff3d590
Commit https://github.com/c3js/c3/commit/ff3d5900f2ecc29fbdf2a0bb40707e23e8150ecd by @kt3k contains a (probably unintentional) breaking change:

Before https://github.com/c3js/c3/commit/ff3d5900f2ecc29fbdf2a0bb40707e23e8150ecd d3 had to be loaded **only before the first ChartInternal is created** (which happens when c3 is first used ).
After https://github.com/c3js/c3/commit/ff3d5900f2ecc29fbdf2a0bb40707e23e8150ecd d3 has to be loaded **before c3.js is first executed.**

The breaking change makes it impossible to load d3 and c3 in parallel which is odd as it breaks all code that relies on that behavior.

The following code is responsible:
```javascript
var d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;
```

Before there was no toplevel d3 object but the d3 referenced in ChartInternal 
```javascript
$$.d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;
```

In my opinion the old logic made much more sense as it allows for optimizing page loading times!

Just assigning the d3 object *first* in the ChartInternal and all should be fine.

kt3k:
fix #2060

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -19,6 +19,7 @@
     "jasmine": true,
 
     "globals": {
-        "d3": false
+        "d3": false,
+        "require": false
     }
 }
--- a/src/core.js
+++ b/src/core.js
@@ -1,4 +1,3 @@
-import * as d3 from 'd3';
 import Axis from './axis';
 import CLASS from './class';
 import { isValue, isFunction, isString, isUndefined, isDefined, ceil10, asHalfPixel, diffDomain, isEmpty, notEmpty, getOption, hasValue, sanitise, getPathBox } from './util';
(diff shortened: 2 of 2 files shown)


aendrew:
LGTM. Merging now because it's breaking things for people otherwise.

PR #2069, opened 2017-06-06
Moves bar and arc opacity styling to CSS; supersedes #1069.

aendrew:
See original PR at #1069.

Original PR was tested and RTBC, and it doesn't seem like anything's really changed since then, but might be worth testing briefly to ensure it still works.


MaxwellM:
I didn't believe I would have permission to approve this, but I was wrong.

kt3k:
Released as [v0.4.15](https://github.com/c3js/c3/releases/tag/v0.4.15).

PR #2070, opened 2017-06-06
Add support for tooltip ordering, fix #1050 #1764 [REPOST PR #1813]

### LINKED ISSUE #1050: Tooltip options

alexgerv:
This is a repost of PR #1813 , on master. 
All code by @panthony.


aendrew:
Thanks @alexgerv! :tada:

kt3k:
Published at v0.4.13.

PR #2083, opened 2017-06-15
Use chrome in testing

kt3k:
This PR changes the testing browser from phantomjs to chrome.

PhantomJS doesn't seem maintained now. ( https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE )
I think we shouldn't use it anymore.

- I relaxed some assertions a little to make it pass in chrome.
- I added circle.yml to update chrome to the latest version
- I added codecov.yml to relax the commit status from codecov.
  - This setting means that it allows 0.1% decrease in PR's coverage diff.


kt3k:
I think ideally we should test on all major browsers. Saucelabs seems one of the ideal solutions, but the tunneling between CI services and Saucelabs seems unstable sometimes to me. If saucelabs is difficult, I think appveyor can be an alternative for IE testing platform.

Thanks for reviewing! 😄

PR #2088, opened 2017-06-24
[Refactor] Write unit tests of convert(Rows|Columns)ToData

kt3k:
- Introduced babel in src/ directory to improve the source quality.
- ~~Moved `convertColumns(Rows)ToData to data.convert.utils` because they are actually static methods and it makes them easy to unit test.~~
- Added unit tests of convert(Rows|Columns)ToData


masayuki0812:
Thanks for the PR. Basically looks great, but I don't agree with the second part about `data.convert.utils` because those methods should be accessible through internal object (e.g. Other internal functions would want to use them through its reference to internal object in order to manipulate data). This is a part of c3-way, so I feel we should keep this feature.

kt3k:
Updated the PR and moved convert functions back to data.convert.js.

masayuki0812:
LGTM!

PR #2096, opened 2017-07-06
added `.inverted` class

ceremcem:
Usage: Add `.inverted` class to `.chart` to make axis' colors inverted.


kt3k:
Talked about this with @masayuki0812, and we think that this should be covered in the documentation and examples, and should not be added as core feature.

We created issue for that: https://github.com/c3js/c3/issues/2109

PR #2099, opened 2017-07-09
Introduce AxisInternal

masayuki0812:
This PR introduces `AxisInternal` and makes it possible to access axis internal functions.


kt3k:
I agree with the structure of this refactoring.

As we talked informally on chat, I prefer using `this` and arrow functions instead of `$$$` convention. (`$$$` seems a bit strange to me.)

src/axis.js
```diff
@@ -1,16 +1,386 @@
...
+        else {
+            if (!arguments.length) { return $$$.tickValues; }
+            $$$.tickValues = x;
+        }
+        return axis;
+    };
+    return axis;
+};
+
+export default function Axis(owner) {
```
kt3k: How about using class keyword like the below?

```js
export default class Axis extends Component {
  constructor (owner) {
    ...
  }
}
```

By the above we don't need to maintain `inherit` util anymore.

masayuki0812: Looks good. Implemented by f1f8ab60a570ae72fd5af005b14289a91e410594 .

masayuki0812:
Fixed `$$$` and changed to use `class` and `extends` instead of `inherit`.

PR #2100, opened 2017-07-11
Log scale with axis y / y2

panthony:
Tried to tackle this with relative success:

https://jsfiddle.net/f2nenq10/6/

Known issues:
 - stacked bar chart
 - padding with data labels

Any help with this would be appreciated.


dmjio:
@panthony any way we can re-open and merge?

PR #2111, opened 2017-07-20
Move dist files to dist/ and ignore built files

### LINKED ISSUE #2076: Move built files to dist, remove from version control
A thought — should we set `lib/index.js` to the `module`, `dist/c3.js` to the `main` attribute in `package.json`, move all the built files to `dist`, put `dist` in `.gitignore` and then set Travis or Circle to build on release?

## Advantages:

* No built files showing up in PRs
* Using C3 as a ES6 module totally bypasses build process — it gets built as part of downstream bundler in consumer project
* This in turn will enable people creating plugins to bundle them and C3 together in downstream projects, which may result yield better file sizes due to letting the bundler handle minification. Possibly, I'm not at all sure.
* Possibly opens the way for better use on the server?

Reference re: `pkg.module`: https://github.com/rollup/rollup/wiki/pkg.module

kt3k:
Many people send pull requsts with built file changes. This PR prevents that situations.

---
closes #2076


panthony:
@kt3k I think this PR will break everyone relying on this project since they'll need to update the location of the included files no?

It should be clearly documented somewhere.

kt3k:
@panthony 
If one require c3 like `var c3 = require('c3')` in webpack/browserify, then it follows `main` field of package.json and doesn't break in this case.

However If one use CDN like rawgit `https://rawgit.com/c3js/c3/master/c3.js`, that breaks. Because of this, this change breaks most of existing fiddles and needs updates on c3's url.

Anyway, as you say, we need to document clearly how this breaks the code before merging.

kt3k:
Closing for now. I think there would be a better solution, which does not move {c3,c3.min}{.js,.css} and `npm run build` does not update them. I'll reopen when that one ready.

Issue #2136, opened 2017-08-11
Horizontal bar stroke doesn't get dynamic color


nbdavies:
I'm trying to style a bunch of bar charts to have opaque edges and translucent interiors:

```
.c3-bar {
  stroke-width: 2px;
  stroke-opacity: 1;
  fill-opacity: 0.8;
}
```

The color of the bars in these charts is determined via a data.color function that's different for each chart.

For vertical bar charts this works great, because the same color (determined via function) gets applied to the bar's fill and its stroke. But for horizontal bar charts, with flipped axes, the color determined by the function only gets applied to the bar's fill, not to its stroke. So the stroke color defaults to black.

Because the color of each bar is being dynamically determined by function, it's not practical to override the colors from a CSS file.

nbdavies:
I think I see what's going on. In c3/shape.bar.js, updateBar() updates both:
```
        .style("stroke", color)
        .style("fill", color);
```
While redrawBar() does not:
```
            .style("fill", this.color)
```
Horizontal bar graphs must be doing a redrawBar(). I tested making redrawBar() consistent with updateBar(), and it would indeed solve the problem.

I'll try to put together a PR.

kt3k:
Thanks for reporting!

Confirmed as a bug. Thanks!
https://jsfiddle.net/kt3k/p2ncr8xs/4/

PR #2137, opened 2017-08-11
Dynamically color stroke when redrawing bars

nbdavies:
Fixes issue https://github.com/c3js/c3/issues/2136 by making redrawBar() consistent with updateBar().


kt3k:
Thanks for the contribution!

PR #2164, opened 2017-09-05
Call resize handlers only if chart is displayed

upphiminn:
In the current implementation the resize handlers are called even if the chart is not displayed. This is not ideal when having one chart in focus and some hidden ones, as all event handlers get called and performance degrades. Also, while `Chrome` behaves correctly and does not actually change the width of the hidden ones, `Firefox` tries to resize the hidden elements and sets incorrect full widths, visible after the hidden charts are re-displayed. By checking the `offsetParent` we can tell if the chart is displayed or not and we can skip resize handling if hidden. 

@kt3k

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -937,11 +937,20 @@ c3_chart_internal_fn.bindResize = function () {
     $$.resizeFunction.add(function () {
         config.onresized.call($$);
     });
+    
+    var resizeIfElementDisplayed = function() {
+        // if element not displayed skip it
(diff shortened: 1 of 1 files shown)


kt3k:
Thanks for the contribution. This change makes sense to me.

`offsetParent` seems to be null as well when the element itself has `position: fixed` style, but I think nobody use `position: fixed` directly to the chart's element. (possibly somebody would use `position: fixed` for the parent of the chart, but that doesn't cause a problem.)

Issue #1790, opened 2016-08-03
Hide/unhide toggle to blanks out chart


achourasia:
When trying to hide/unhide lines, the rendering fails and the chart blanks out
The data has very large and small numbers, is something running out of range?
http://jsfiddle.net/swwg3ceh/

I also tried to add the following in data, which also blanks out the chart
hide: ['M0', 'area',],

Nice library! thanks

achourasia:
It seems like c3js does not like column names to have numerals in it 'M0'

PR #2172, opened 2017-09-14
make the target completely hidden

Xuefeng-Zhu:
Fix https://github.com/c3js/c3/issues/1790


panthony:
Hi,

Do you think you could add some asserts for this in 'api.show-spec'?

Thanks!

Xuefeng-Zhu:
@panthony Sorry, we may not have time to do it at this point

panthony:
@Xuefeng-Zhu No worries, hard to ask for tests so long after :)

panthony:
Tested and it does fix the issue.

PR #2192, opened 2017-10-10
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that c3 gets 69M hits per month on jsDelivr, so I thought you might like this badge.


kt3k:
Thanks! 😄

PR #2232, opened 2017-11-27
add legend format

pkpatels:


c3.js
```diff
@@ -1,7 +1,7 @@
 (function (global, factory) {
-	typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
-	typeof define === 'function' && define.amd ? define(factory) :
-	(global.c3 = factory());
+    typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
```
panthony: You modified the distribution file!

You should modify the proper file from src directory and not commit the c3.js file which is overwritten during release

c3.js
```diff
@@ -5097,6 +5097,7 @@ c3_chart_internal_fn.getDefaultConfig = function () {
         legend_padding: 0,
         legend_item_tile_width: 10,
         legend_item_tile_height: 10,
+        legend_format: undefined,
```
panthony: Please, also add the documentation for this new option. The documentation site has been added to the repository.

c3.js
```diff
@@ -7445,8 +7450,12 @@ c3_chart_internal_fn.updateLegend = function (targetIds, options, transitions) {
         background = $$.legend.insert('g', '.' + CLASS.legendItem).attr("class", CLASS.legendBackground).append('rect');
     }
 
-    texts = $$.legend.selectAll('text').data(targetIds).text(function (id) {
-        return isDefined(config.data_names[id]) ? config.data_names[id] : id;
+    texts = $$.legend.selectAll('text').data(targetIds).text(function (id, index) {
+        if(config.legend_format && isFunction(config.legend_format)) {
+            return config.legend_format(id, index);
+        } else {
+            return isDefined(config.data_names[id]) ? config.data_names[id] : id;
```
panthony: Please, do not repeat the logic that resolve the legend's name.

panthony (changes requested):
Thanks for your contribution.

Please address the requested changes so that we can proceed.

c3.js
pkpatels: I think this is just the indentation issue with my editor. I don't see any chages here.

c3.js
pkpatels: I didn't get you here.

panthony: @pkpatels I saw the same logic twice but it must have been between c3.js and legend.js files.

Your pull request still includes the `c3.js` file although it should not be committed.

c3.js
panthony: What I meant is that you modified the `c3.js` file which is the distribution file and that this file should not be part of the pull request.

pkpatels: @panthony i have make the requested changes. I have moved the logic to legend.js and config.js

You can review chnages.

pkpatels:
@panthony I have made the requested changes.

PR #2203, opened 2017-10-23
D3 v3/4 compatibility

mihailik:
Basic work for #2143

Divergent D3 dependencies are extracted into functions alongside this naming style: `$$.d3_newName` (where newName is the name of the function in v4).

Three problematic places currently:

* <a href="https://github.com/d3/d3/blob/master/CHANGES.md#time-formats-d3-time-format">Time formats</a> (fails at core.js around defaultAxisTimeFormat definition)
* <a href="https://github.com/d3/d3/blob/master/CHANGES.md#brushes-d3-brush">d3.brush</a> and <a href="https://github.com/d3/d3/blob/master/CHANGES.md#zooming-d3-zoom">d3.zoom</a> look a bit more complex than just rename

PR #2246, opened 2018-01-01
Update for D3 v4

### LINKED ISSUE #2143: Migrating to d3 v4
### TODO

- [x] enumerate d3 deps of c3.
  - https://gist.github.com/kt3k/8f4914a4d4b9e772cf03b6a0b1ac07f2
- [ ] create mapping of d3.v3 usages to d3.v4 apis.

|  | d3.v3 | d3.v4 |
|-|-------|------|
| d3-selection | d3.select(selector) | d3.select(selector) |
| | d3.selectAll(selector) | d3.selectAll(selector) |
| d3-transition | d3.transition() | d3.transition() |
| d3-array | d3.max(array) | d3.max(array) |
| | d3.min(array) | d3.min(array) |
| | d3.merge(array) | d3.merge(array) |
| d3-scale | d3.scale.category10() | d3.scaleOrdinal(d3.schemeCategory10) |
| d3-scale | d3.scale.linear() | |
| d3-shape | d3.layout.pie() | d3.pie() |
| | d3.svg.arc() | d3.arc() |
| d3-interpolate | d3.interpolate(a, b) | d3.interpolate(a, b) |
| ? | d3.time | |
|  | d3.time.format | |
|  | d3.time.format.utc | |

- [ ] determine the necessary range of d3.v4 apis.
- [ ] rewrite code

### DONE
---
refs #2007, #1648

masayuki0812:
I just updated c3 to work with d3 v4.

Breaking changes I committed:
- Disabled drag in zoom when `data.selection.draggable = true` because of new d3 drag behavior. (zoom and data.selection.draggable still work together, but zoomed section cannot be moved by drag.)
- Disabled multiple event rects because of new d3 zoom behavior. (This could be avoided but only one event rect is very efficient, so this change seems not bad in my opinion.)
- Changed internal functions' interface to accept transitions because of new d3 transition behavior. (Transition is not inherited within child selections in v4, so transition object needs to be passed to each draw function.)
- Renamed `axis.x.extent` to `axis.x.selection` based on d3.v4 update.
- Renamed `zoom.extent` to `zoom.initialRange` based on d3.v4 update.
- Changed `api.tooltip.show` interface. (Removed `index` and added `id` to specify target of data)
- Changed `axis.y.tick.time.value` to `axis.y.tick.time.type` and to receive d3 interval functions (e.g. `d3.timeSecond`).

#2203 and my changes should be the one later.

---
closes #2143


pchaganti:
👍

tgaff:
Not to nag but, d3 is on RC3 of v5 now.  https://github.com/d3/d3/releases

nsmgr8:
d3 v5 doesn't seem to have v4 incompatibility issues.

Very much waiting for this to be merged and released!

dborstelmann:
Now using this feature branch in a production project because Salesforce Lockerservice breaks with d3v3.  Would love to use the actual release version!  Thanks.

dborstelmann:
@masayuki0812 We are noticing in error on tooltip hovers in this branch.

```[Cannot read property 'getItem' of undefined]
ChartInternal.c3_chart_internal_fn.isWithinBar()```

Have you seen this before or do we need to fix it ourselves by hacking it together for the time being?

**Comment by wuno** [comment]
@masayuki0812 We noticed that this problem had to do with  `SVGPathSegList`  not being in scope anymore once the component loaded in Sales Force. I realize this was not made with Sales Force lightning's Locker Service in mind but we were able to come up with a solution. 

I am wondering what your opinion is on this solution and if you can suggest a way we can make it better or solve our problem all together. 

Since the specific function failing was `isWithinBar` we added these 2 lines to bring `SVGPathSegList`   back in scope.

```
var seg0 = new window.SVGPathSegList(that).getItem(0);
var seg1 = new window.SVGPathSegList(that).getItem(1);
```

This is what the function looks like now,

```
c3_chart_internal_fn.isWithinBar = function (mouse, that) {
    var seg0 = new window.SVGPathSegList(that).getItem(0);
    var seg1 = new window.SVGPathSegList(that).getItem(1);
    var box = that.getBoundingClientRect(),
        x = Math.min(seg0.x, seg1.x),
        y = Math.min(seg0.y, seg1.y),
        w = box.width,
        h = box.height,
        offset = 2,
        sx = x - offset,
        ex = x + w + offset,
        sy = y + h + offset,
        ey = y - offset;

    return sx < mouse[0] && mouse[0] < ex && ey < mouse[1] && mouse[1] < sy;
};
```

We are confident that this solves the immediate problem of hovers on bar graphs showing up, but the concern is what else is being affected due to `SVGPathSegList` not being in scope. 

If you can think of a good way for us to solve this throughout your C3.js PR based on the problem and the way we solved the problem, please let us know.
@dborstelmann

masayuki0812:
@dborstelmann @wuno 
Thank you for reporting, but can you open a new issue about `SVGPathSegList` in SalesForce? I don't think it's not related to this update and I'm not going to fix that in this PR.

PR #2272, opened 2018-01-29
Major memory leak fix

### LINKED ISSUE #926: Repeatedly calling c3.generate leaks memory in 0.4.9

nazihahmed:
the detach event listener was not detached properly after calling .destroy
so I basically added resizeIfElementDisplayed to $$ (this), and detach the event listener properly
also .remove must be called on $$.resizeFunction on destroy,
anyways this fixed all my memory issue,
Fix #926


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "c3",
+  "name": "alooma-c3",
```
aendrew: This will break everything if merged into master.

nazihahmed: i know i did this to publish to npm, i will change into a specific branch for merging for you

nazihahmed:
https://github.com/c3js/c3/pull/2273

PR #2273, opened 2018-01-29
Memory leak fix

### LINKED ISSUE #926: Repeatedly calling c3.generate leaks memory in 0.4.9

nazihahmed:
the detach event listener was not detached properly after calling .destroy
so I basically added resizeIfElementDisplayed to $$ (this), and detach the event listener properly
also .remove must be called on $$.resizeFunction on destroy,
anyways this fixed all my memory issue,
Fix #926


nazihahmed:
we are waiting for the merge to happen so we can update our code base with yours,
pls release a new version to npm after these changes

package.json
```diff
@@ -60,8 +60,7 @@
     "node-sass": "^4.5.3",
     "node-static": "^0.7.9",
     "nodemon": "^1.11.0",
-    "rollup": "^0.55.0",
```
kt3k: I think `rollup` is necessary to build the bundled version (c3.js).

.babelrc
```diff
@@ -0,0 +1,13 @@
+{
```
kt3k: Because babel is configured in `rollup.config.js` (when bundling) and `karma.conf.js` (when testing), I think this file is not necessary and confusing.

kt3k:
Thanks for finding out these! 👍

package.json
nazihahmed: It must be global because in the scripts where it's called you are using the global rollup command not from node modules, anyways i will add it back

.babelrc
nazihahmed: For me the build keeps returning an error from rollup until i added this file

kt3k: Do you use `npm run` command for invoking rollup?

`npm run build` command should invoke rollup (and other tools) with correct configs, like:

```console
$ npm run build

> c3@0.4.18 build /Users/kt3k/t/c3
> npm run build:js && npm run build:css


> c3@0.4.18 build:js /Users/kt3k/t/c3
> npm run build:js:rollup && npm run build:js:uglify


> c3@0.4.18 build:js:rollup /Users/kt3k/t/c3
> rollup -c > c3.js

(!) Some options have been renamed
https://gist.github.com/Rich-Harris/d472c50732dab03efeb37472b08a3f32
entry is now input
moduleName is now output.name
format is now output.format
...
```

`npm run` command automatically uses local rollup cli and you don't need the global one.

nazihahmed: yes, I do use this, but the issue is not rollup, rollup works but throws a different error (It looks like your Babel configuration specifies a module transformer), so I added this file and it worked for me, anyways I removed this file for you to merge.
when will you publish the next version to npm? we are still using our own library.

Mobiletainment:
is there any update on this? Would love to see it go into master.

nazihahmed:
@Mobiletainment you can use [alooma-c3](https://www.npmjs.com/package/alooma-c3) while they release the new version, after they release it you simply change back to c3 in your package.json

kt3k:
The root problem was that the attached function (resizeIfElementDisplayed) and the detached one (resizeFunction) were different and the attached one was leaked every time. This seems to have started at #2164. Thanks again for finding out this!

By the way $$.resizeFunction.remove() doesn't seem removing internal resize functions, because it removes the only given function (see https://github.com/c3js/c3/blob/7e5c20c0ffd098ad7a8fc437f19ccea069d60edf/src/core.js#L988-L995). We follow up this on another PR.

kt3k:
https://github.com/c3js/c3/releases/tag/v0.4.19

PR #1355, opened 2015-08-28
multi arc gauge enhancement to draw more than one arc in gauges

GoekhanOezen:
See issue #1077.

For each `data_column` one arc, the related background arc, arc label and a line to highlight the relation from arc to label will be drawn. When there are multiple `data_column` the legend is shown. For cleaner calculation in functions some variables are added. When there only is one `data_column` visible the old behaviour didnt changed.

This is my first pull request. We are already using the feature. I spend much time to refactor but i am sure there are still some optimiziations left.

All existing tests passed.
I would be glad to hear your opinion to the changes.

multi arc gauge 1
multi arc gauge 2
multi arc gauge 3

PR #2278, opened 2018-02-01
Follow up `multi arc gauge` PR (#1355)

kt3k:
This follows up #1355 

- In the original test setup, `type: 'gauge'` was set in a wrong place. I fixed it and it broke several test cases. I fixed them as well by changing selector, expected values etc. See https://github.com/c3js/c3/commit/acb6810e5929c714dd5b4bcc720191dad8017a5e for details.
- I added the example under `htdocs` which renders like the below:

### DIFF
--- a/htdocs/index.html
+++ b/htdocs/index.html
@@ -76,6 +76,9 @@
               <a href="./samples/chart_gauge.html">
                 Gauge chart with ordinary data
               </a>
+              <a href="./samples/chart_multi_arc_gauge.html">
+                Gauge chart with multiple arcs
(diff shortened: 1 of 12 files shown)


srinivasyl:
@kt3k can we provide two different max values for each data source like you've padded1, padded2, padded3, and padded. I want to set 100 as max value for padded1 and 200 as max value for padded2 and so on is that possbile in c3.js

surbhibh:
@kt3k i am not able to implement multiArc Gauge. Can you please help me with this.

PR #2279, opened 2018-02-01
use shield circleci badge for consistency with other badges

panthony:
This PR switch the default circleci badge:

[CircleCI](https://circleci.com/gh/c3js/c3)

By the shield style:

[CircleCI](https://circleci.com/gh/c3js/c3)

Which is much more consistent with other badges.


kt3k:
I didn't know that option exists! Thanks! 👍

PR #2282, opened 2018-02-01
Makes unfilled portion of 360 gauge visible

### LINKED ISSUE #1602: Gauge unfilled section for full circle is not visible.
If the gauge is set to fullCircle and has a data value that is less than the max value, the unfilled section of the gauge is not visible.

Kinlaw:
See JSFiddle http://jsfiddle.net/kinlaw/vkt2pvc9/1/ for example of issue.

Fixes #1602


kt3k:
Released as [v0.4.20](https://github.com/c3js/c3/releases/tag/v0.4.20). Thanks for the contribution! @Kinlaw

Kinlaw:
Awesome! Thanks for merging!

Issue #2251, opened 2018-01-05
c3 rounds negative values on the x axis


ashleydavis:
I've found that C3 rounds negative value on the x axis, it doesn't have this behavior for positive values.

It should function the same way for both negative and positive values displayed on the x axis.

C3 version: 0.4.18
D3 version: 3.5.14
Browser: Chrome  63.0.3239.84 
OS: Windows 10

Here is a screenshot that shows the problem:

![c3 issue - negative values on the x axis are rounded](https://user-images.githubusercontent.com/625126/34593578-6e838054-f217-11e7-80fd-510952e1e572.png)

Jsfiddle example that demonstrates the problem:

https://jsfiddle.net/ee0Lxv07/

panthony:
Fixed by 81316cd

ashleydavis:
Thanks. C3 is awesome. I've written a chapter on it in my upcoming book!

http://bit.ly/2t2cJu2

PR #2297, opened 2018-02-17
Issues/2251

muttox:
potential fix and tests for #2251.

Tracked getXAxisTickFormat as far back as 1987be65cf7eef5666b191c11cf184e67a3cf5df but couldn't find an explanation for why the change from returning a null value needed to truncate on values < 0, so removed that portion and created matching testcase.

### DIFF
--- a/spec/axis-spec.js
+++ b/spec/axis-spec.js
@@ -153,6 +153,25 @@ describe('c3 chart axis', function () {
     });
 
     describe('axis.x.tick.values', function () {
+        describe('formatted correctly when negative', function() {
+            var xValues = [-3.3, -2.2, -1.1, 1.1, 2.2, 3.3];
+            beforeEach(function() {
+                args.data = {
+                    x: 'x',
+                    columns: [
+                        ['x'].concat(xValues),
+                        ['data1', 30, 200, 100, 400, 150, 250],
+                    ]
+                };
+            });
+
+            it ('should not generate whole number for negative values', function() {
+                var tickValues = [];
+                d3.select('.c3-axis-x').selectAll('g.tick').selectAll('tspan').each(function(d, i) { expect(tickValues.push(parseFloat(d.splitted)) === xValues[i]); });
+
+            });
+        });
+
         describe('function is provided', function () {
             var tickGenerator = function () {
                 var values = [];
--- a/src/axis.js
+++ b/src/axis.js
@@ -475,8 +475,15 @@ c3_axis_fn.getId = function getId(id) {
     return id in config.data_axes ? config.data_axes[id] : 'y';
 };
 c3_axis_fn.getXAxisTickFormat = function getXAxisTickFormat() {
-    var $$ = this.owner, config = $$.config,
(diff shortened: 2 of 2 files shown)


src/axis.js
```diff
@@ -472,8 +472,10 @@ c3_axis_fn.getId = function getId(id) {
     return id in config.data_axes ? config.data_axes[id] : 'y';
 };
 c3_axis_fn.getXAxisTickFormat = function getXAxisTickFormat() {
+    // #2251 previosly set any negative values to a whole number,
```
panthony: previously*

src/axis.js
```diff
@@ -472,8 +472,10 @@ c3_axis_fn.getId = function getId(id) {
     return id in config.data_axes ? config.data_axes[id] : 'y';
 };
 c3_axis_fn.getXAxisTickFormat = function getXAxisTickFormat() {
+    // #2251 previosly set any negative values to a whole number, 
+    // however both should be truncated according to the users format specification
     var $$ = this.owner, config = $$.config,
-        format = $$.isTimeSeries() ? $$.defaultAxisTimeFormat : $$.isCategorized() ? $$.categoryName : function (v) { return v < 0 ? v.toFixed(0) : v; };
+        format = $$.isTimeSeries() ? $$.defaultAxisTimeFormat : $$.isCategorized() ? $$.categoryName : function (v) { return v; };
```
panthony: I'm really bored of those double ternary that are hard to read!

Why not:

```js
let format;
if ($$.isTimeSeries()) {
  format - $$.defaultAxisTimeFormat;
} else if ($$.isCategorized()) {
  format = $$.categoryName;
} else {
  format = (v) => v;
}
```
@kt3k I know this is not the current coding style of c3js but why make it hard on us understand/review ? What's your view on this?

muttox: @panthony an alternative if you wanted to keep the ternary style but clear up the formatting could be:
```javascript
let format = ($$.isTimeSeries()) 
        ? $$.defaultAxisTimeFormat 
        : ($$.isCategorized()) 
            ? $$.categoryName 
            : function (v) { return v; };
```

panthony: @muttox Or this way yes. That's way easier to read IMHO and at least when you modify a code path it does not highlight the whole thing.

spec/axis-spec.js
```diff
@@ -153,6 +153,32 @@ describe('c3 chart axis', function () {
...
+                };
+            });
+
+            it ('should not generate whole number for negative values', function() {
+                var tickValues = [];
+                d3.select('.c3-axis-x').selectAll('g.tick').selectAll('tspan').each(function(d, i) { expect(tickValues.push(parseFloat(d.splitted)) === xValues[i]); });
+
+            });
+
+            // it ('should not generate whole number for negative values', function() {
```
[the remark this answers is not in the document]
muttox: test is unnecessary, was an initial attempt used as reference, fixed the other spelling errors in comments as well as rebased to upstream and squashed.

Issue #1386, opened 2015-09-18
Time series graph, overlapping bars


vishal-kony:
Why the bars of my timeseries graph are overlapping? 

![2015-09-18_1208](https://cloud.githubusercontent.com/assets/10286364/9956895/c1ad6842-5e19-11e5-869a-c85805cd6254.png)

here's my code.. 

var chart = c3.generate({
    bindto: '#chart',
    data: {
        x: 'date',
        xFormat: '%m/%d/%y %I:%M %p',
        json: final_data,
        keys: {
            x: 'date',
            value: values
        },
        type: 'bar',
        groups: val
    },
    axis: {
        x: {
            type: 'timeseries',
            tick: {
                format: '%b %d',
                centered: true,
                fit: true
            }
        }
    },
    zoom: {
        enabled: true
    },
    color: {
        pattern: colors
    }
});

panthony:
Very likely a duplicate of #1329

Codewolf:
Sorry to resurrect a dead (albeit open) ticket, but is there any update on this as is still an issue on the latest stable version (c3 0.4.15)
![image](https://user-images.githubusercontent.com/234298/28679083-04274e0e-72ea-11e7-8391-2625bceb5a9d.png)

panthony:
@Codewolf I don't think so, a work around is to fill the gaps with null values.

Codewolf:
so, to clarify: i would need to programmatically (either client or server side) identify the bounds (datetime upper and lower) and manually fill in any date gaps with a 'null' or '0' value?

Surely that would then display the dates with a value of 0 which isn't correct as nothing (value wise) actually happened on those dates?

panthony:
Yes, find the bounds and fill the gaps with either null or 0 depending on what you are displaying, 0 does not always means null.

Before:
https://jsfiddle.net/mcd6ucge/31/

After:
https://jsfiddle.net/mcd6ucge/306/

Note that if you fill with 'null', yes the date is displayed on X axis, but no, they are no displayed with a value of 0, there is nothing on hover for this date.

Codewolf:
I see, the main issue with this is if a large date range is created (for example 1 year) and only a few of those dates have values i would end up with a (mostly) empty graph and unreadable dates on the X axis coupled with very thin (~1 or 2 px depending on graph width) bars.

panthony:
Regardless of this bug, the date seems to be evenly distributed on the X axis so it would not be any different. You would still have large empty space between each bars.

[...]

PR #2306, opened 2018-03-14
Fix bar overlap bug when tick fit is false

rkrabek:
Partial fix for issue #1386 

If tick fit is set to false and there are many values, the bars overlap because the bar width is currently calculated based on how many ticks there are instead of the data points.

### DIFF
--- a/spec/shape.bar-spec.js
+++ b/spec/shape.bar-spec.js
@@ -97,7 +97,6 @@ describe('c3 chart shape bar', function () {
                 });
            });
         });
-
     });
 
     describe('internal.isWithinBar', function () {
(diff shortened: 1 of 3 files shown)


panthony:
@rkrabek Thanks for your contribution. Could you add some to tests to ensure this is actually fixing something?

I tried your fix with the following config:

http://jsfiddle.net/moeumaj7/3/

And in this case it's actually break the display.

rkrabek:
@panthony Sorry for the delayed response, I've been super busy at work. I've updated the PR to include tests for bar width when tick fit is false. 

The config that you provided: http://jsfiddle.net/moeumaj7/3/ fails with my patch for the same reason it fails when you switch tick/fit to true as shown here: http://jsfiddle.net/ox76hL53/
For the sake of not introducing regressions, I've updated the code to take the max of ticks or unique x values so that should address the break in the config you provided. 

This patch aims to fix this scenario (with or without the null data points): http://jsfiddle.net/dtttsa9d/

PR #1814, opened 2016-08-22
Pie and donuts really handle data_order as advertised by documentation.

panthony:
Although the `data_order` [reference](http://c3js.org/reference.html#data-order) state that it is also used to sort the pieces of the pie chart, this is not the case.

This PR will fix this (see #1159).

PR #2311, opened 2018-03-21
Fix tooltip randomly switching when have equal values

### LINKED ISSUE #2293: Tooltip legend randomly switching order of equal values
Hi,

I have a stacked bar chart, and I chose the legend order to the descent. However, when there are 2 or more equal values, the legend order shown on mouseover tooltip keeps randomly changing the order of the values that are equal.

Here you can see an example: https://jsfiddle.net/cgwLqbL8/ . Just pass the mouse over the bar and look at the tooltip order.

Is there any way to avoid this?

panthony:
This PR fixes issue described by #2293 and introduced by #1814.

Hopefully, this time, it does not introduce any more regressions.

There are lot's of tooltip/donut tests around sorting and they are all greens.

---
close #2293

### DIFF
--- a/src/arc.js
+++ b/src/arc.js
@@ -7,7 +7,16 @@ c3_chart_internal_fn.initPie = function () {
     $$.pie = d3.layout.pie().value(function (d) {
         return d.values.reduce(function (a, b) { return a + b.value; }, 0);
     });
-    $$.pie.sort($$.getOrderFunction() || null);
+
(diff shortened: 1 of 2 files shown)


kt3k:
I confirmed this fixes the example of #2293. Thank you for the contribution!

(The iteration of sorting and reversing caused the flipping the order of the entries of the same values. This PR removes the reversing of targets and flipping behavior disappears.)

PR #2314, opened 2018-03-22
Fix targets ordering when containing numerical value as name

### LINKED ISSUE #1534: Wrong stack order when names a mix of text only and numbers only
The below code breaks the order and puts the number only name first 

```js
    data: {
        columns: [
            ['data1', 130, 200, 320, 400, 530, 750],
            ['2', 130, 10, 130, 200, 150, 250],
            ['data3', 130, 50, 10, 200, 250, 150]
        ],
        type: 'bar',
        groups: [
            ['data1', '2', 'data3']
        ] 
   }
```

image
This code works as expected

```js
    data: {
        columns: [
            ['data1', 130, 200, 320, 400, 530, 750],
            ['2data', 130, 10, 130, 200, 150, 250],
            ['data3', 130, 50, 10, 200, 250, 150]
        ],
        type: 'bar',
        groups: [
            ['data1', '2data', 'data3']
        ] 
   }
```

image

### LINKED ISSUE #1793: convertColumnsData do not preserve order when a category is '0'
With c3js `0.4.11` and probably the latest, the `data.order: null` is not honored correctly when one of the category is '0'.

With the following configuration:

``` javascript
var chart = c3.generate({
    data: {
        columns: [
            ['< 0', 1],
            ['0', 2],
            ['> 0', 3]
        ],
        type : 'pie',
        order: null 
    },
    color: {
      pattern: ['green', 'grey', 'red']
    }
});
```

I would except to have the slices in the following order `< 0`, `0` then `> 0` whereas I got `0`, `< 0` then `> 0`.

See: https://jsfiddle.net/mcd6ucge/126/

panthony:
This PR aim to fix issue where `data.order: null` would still not leave the data order untouched if targets have numerical values.

- Fix #1793 
- Fix #1534

The idea of this update is to keep the ordered keys as an array in the pivot format returned by convert methods.

### DIFF
--- a/src/data.convert.js
+++ b/src/data.convert.js
@@ -1,5 +1,5 @@
 import { c3_chart_internal_fn } from './core';
-import { isValue, isUndefined, isDefined, notEmpty } from './util';
+import { isValue, isUndefined, isDefined, notEmpty, isArray } from './util';
 
 c3_chart_internal_fn.convertUrlToData = function (url, mimeType, headers, keys, done) {
     var $$ = this, type = mimeType ? mimeType : 'csv';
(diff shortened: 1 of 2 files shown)


src/data.convert.js
```diff
@@ -131,16 +130,25 @@ c3_chart_internal_fn.convertColumnsToData = (columns) => {
...
             newRows[j - 1][key] = columns[i][j];
         }
+        keys.push(key);
     }
 
-    return newRows;
+    return { keys, rows: newRows };
 };
 
 c3_chart_internal_fn.convertDataToTargets = function (data, appendXs) {
```
kt3k: Now the shape of `data` is changed to `{ keys, rows }`. I think we should document that in jsdoc format like:

```
/**
 * Converts the data format into the target format.
 * @param {Array} data.keys ...describe...
 * @param {Array} data.rows ...describe...
 * @param {boolean} appendXs
 */
```

kt3k:
I confirmed the data order preseved in this example with this change:

```
var chart = c3.generate({
    data: {
        columns: [
            ['data5', 30],
            ['data1', 130],
            ['2', 35],
            ['data4', 60],
            ['data3', 100]
        ],  
        type: 'bar'
    }   
});
```

Thanks for the contribution!

Issue #1794, opened 2016-08-05
Donut/Pie issue with shown active arcs with some category


panthony:
Not sure if it's the same issue as #1793 but with the following configuration:

``` javascript
var chart = c3.generate({
    data: {
        columns: [
            ['> 0', 1],
            ['== 0', 2],
            ['< 0', 3]
        ],
        type : 'donut',
        order: null,
        onclick: function (d) { console.log('click', d); },
        onmouseover: function (d) { console.log('over', d); }
    },
    color: {
      pattern: ['green', 'grey', 'red']
    }
});
```

If I hover the arc corresponding to `> 0`, the arc for `< 0` is also shown as active.

Works fine if categories are 'a', 'b' or 'c' ...

The over/click events and the tooltip are ok.

See https://jsfiddle.net/mcd6ucge/131/

panthony:
It's because the active/hover state is based on the css class of the arc.

This class is generated by normalizing the category name.

The categories `> 0` and `< 0` are both normalized as `c4-arcs--0` (`>` and  `<` are replaced by `-`).

panthony:
Duplicate of #1687

Issue #1083, opened 2015-03-26
Problem in hiding parts of pie chart when clicked on legend.


Dhruva2015:
I generated pie chart using the data of blood group as shown below. It has +ve and -ve symbols in it. It renders chart properly but when clicked on legend like AB+ve both AB+ve and AB-ve will get hide and chart also will not join properly.

Sample data : 

[[O+ve, 861.0], [B+ve, 1134.0], [O-ve, 270.0], [AB+ve, 279.0], [0+ve, 211.0], [A+ve, 854.0], [AB-ve, 288.0], [A-ve, 1153.0]]

Initial pie chart display on load
![piechart](https://cloud.githubusercontent.com/assets/11553557/6847858/96192504-d3ef-11e4-8226-9f60e65f4a05.jpg)

On click of legend part of pie chart is not hidden
![piechartissue1](https://cloud.githubusercontent.com/assets/11553557/6847863/a0ac5a72-d3ef-11e4-8460-d47b4688ed3c.jpg)

On mouse over both AB+ve and AB-ve parts are highlighted
![piechartissue2](https://cloud.githubusercontent.com/assets/11553557/6847866/a79d6862-d3ef-11e4-9881-aa0e35f2f31c.jpg)

michaelrodov:
Hi
Can you please post JSFiddle or something similar of the reproduction?
Many Thanks!

panthony:
Duplicate with #1687

This is the `+` and `-` sign in `AB+ve` the issue.

PR #2315, opened 2018-03-22
Fix hovering state when target contains special characters

### LINKED ISSUE #1687: When the column name includes sign, C3 cannot tell the difference between signs
**C3 Version:** v0.4.11
**Browsers:** Chrome(Version 50.0.2661.94 (64-bit))   Safari(Version 9.1)
*_Bug: *_ One column name is with '+' and the other is '-', when I only select 'data+' from legend, it will highlight both.

screen shot 2016-05-06 at 1 47 02 pm

panthony:
This PRs fixes the following issue:

- #1794
- #1687
- #1083 (probably)

CSS class names are NOT allowed spaces but every other characters is fine (https://www.w3.org/TR/CSS21/syndata.html#characters).

Since they are all prefixed by `c3-targets-` we do not have to care about the first character.

We simply need to escape in `querySelector` or in CSS if any.

I don't know if `c3-target-XXX`  is considered public or private api but this PR will break css/js based on this css class when target contains special characters.

---
close #1687

### DIFF
--- a/spec/class-spec.js
+++ b/spec/class-spec.js
@@ -17,32 +17,42 @@ describe('c3 chart class', function () {
         chart = window.initChart(chart, args, done);
     });
 
-    describe('internal.getTargetSelectorSuffix', function () {
+    describe('internal.generateTargetClass', function () {
 
         it('should not replace any characters', function () {
             var input = 'data1',
                 expected = '-' + input,
-                suffix = chart.internal.getTargetSelectorSuffix(input);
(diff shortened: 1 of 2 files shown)


kt3k:
I confirmed this fixes the example of #1687 

This change breaks the case when the data name contains special characters other than `-` and the user uses `c3-target-XXX` directly from their code. I think that's very rare usage. (possibly none)

The current replacing rule has an obvious defect, so I think nobody wants that behavior and it's ok to merge this.

PR #2332, opened 2018-03-31
Migrate c3js.org to c3 repo

### LINKED ISSUE #2291: Move the contents of c3js/c3js.org repository here
- The separation of docs and source code makes the maintainance of the project hard.
- Let's move the contents of https://github.com/c3js/c3js.org here!

At the same time, I'd prefer dropping the use of middleman in building of the site. Maybe that's possible by the use of gulp or something similar.

kt3k:
- This PR migrates all the contents of https://github.com/c3js/c3js.org to this repository.
- Defines publish-docs job in circle config and triggers it only when tags are push (= when the new versions are released) to the github.

close #2291


kt3k:
This doesn't seem breaking at least `test` job and therefore it doesn't include serious regression in ci workflow. I'm going to merge this and see if this triggers a build when a tag is pushed.

Issue #2354, opened 2018-04-20
Tooltips and Mouse events not triggering on pie chart


alanhamlett:
Tooltips and Mouse [hover and click events](http://c3js.org/reference.html#data-onclick) aren't working on pie charts.

For example, the official pie chart example from docs doesn't show any console logs from mouse events:
http://c3js.org/samples/chart_pie.html

* **C3 version**: 0.5.3
* **D3 version**: 4.13.0
* **Browser**: Chrome 66.0.3359.81
* **OS**: Mac OS X

alanhamlett:
Adding this CSS/Sass fixed the mouse events and tooltips:

```sass
g.c3-chart {
  g.c3-event-rects {
    rect.c3-event-rect {
      pointer-events: none;
    }
  }
}
```

prathik457:
Can we merge this PR, if this works?

alpeshv:
Adding this css breaks tooltip or bar chart.

alanhamlett:
> Adding this css breaks tooltips on bar chart.

Yea, a better solution is needed to work for all c3 charts.

alpeshv:
As @kvcrawford  has suggested, the issue is `.c3-event-rects` getting rendered as last element. His PR should fix this issue for all the charts. PR got approved. Not sure when its going to be merged.

1Jesper1:
I would like to see this fixed.

kt3k:
Hello @1Jesper1 
We've just published the fix as [v0.6.4](https://github.com/c3js/c3/releases/tag/v0.6.4)! Thanks!

PR #2359, opened 2018-04-28
Append mouse event nodes before chart elements

### LINKED ISSUE #2329: Onclick isn't working for pie charts in C3 version 0.5.1
* **C3 version**:v0.5.1
* **D3 version**:4.13.0
* **Browser**:Chrome
* **OS**:MacOS

When using utilizing Pie Charts, the behavior seems to have changed.  Currently, onclick, mouseover, and mouseout events added to the data element are not working.  It also isn't working in the example at http://c3js.org/samples/chart_pie.html.

code snippet:

```
data:{
columns: data,
type: 'pie',
onclick:function(d,element){
console.log("Clicked");
}
}
```

kvcrawford:
Mouse event nodes were appended last, which covered up the chart elements so that their events weren't firing correctly.

Fixes #2329 and #2354

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -272,6 +272,9 @@ c3_chart_internal_fn.initWithData = function (data) {
     // Grid lines
     if (config.grid_lines_front) { $$.initGridLines(); }
 
+    // Cover whole with rects for events
+    $$.initEventRect();
+
     // Define g for chart
     $$.initChartElements();
 
@@ -280,9 +283,6 @@ c3_chart_internal_fn.initWithData = function (data) {
 
     // Set targets
     $$.updateTargets($$.data.targets);
-
-    // Cover whole with rects for events
-    $$.initEventRect();
 
     // Set default extent if defined
     if (config.axis_x_selection) { $$.brush.selectionAsValue($$.getDefaultSelection()); }


kvcrawford:
Rebased upstream master to stay up to date.

rahul-winner:
When are you guys planning to merge this PR? Is there any regression or functional case being investigated? 
This is an important issue and many are waiting for it. Any update be appreciated. 
Thanks !

kt3k:
@panthony Can you check this change?

panthony:
@kt3k @kvcrawford This PR does fix the events for the pie chart but it breaks the zoom feature (tried with `zoom_category` sample).

Not very familiar with this part not sure what the proper fix could be.

kt3k:
@panthony Thanks! Very helpful!

rahul-winner:
@panthony 
I carried out couple of tests after moving 'initEventRect' before 'initChartElements' as in this PR. I've added following properties for zoom, and I am able to zoom in-out with and without sub chart. I've used C3JS version 0.6.3 and D3JS version 5.5.0.
I tested with and without 'subchart' both.
```
zoom: {
     enabled: true
},
subchart: {
     show: true
}
```
I might have missed some use case as I'm still acquainting to C3JS. Could you please elaborate the use case which might got broken with this change? I'll like to try that at my end. 

Regards, Rahul

rahul-winner:
Thanks @panthony 
I'm not able to build the doc site. I'm not much familiar with Ruby. I've listed the error I'm getting [here](https://gist.github.com/rahul-winner/951ed0f3706110cc91e46bd1531002bd). Could you please take a look and help me out. 
Or please let me if there any other way to run the samples.

panthony:
@rahul-winner Unfortunately I never took a look at the actual code that create the doc site and the fact that you are on windows is not helping I'm not familiar with the whole node environment on this platform.

This is using middleman at the moment, there is a section about Windows user:

https://github.com/middleman/middleman#installation

Maybe you are missing a dependency ?

The weird thing is that the sample that is not working for me pretty much only do

```js
zoom: {
     enabled: true
},
subchart: {
     show: true
}
```

Which you claimed worked... :/

rahul-winner:
Thanks @panthony let me check that middleman page.
Meanwhile, @kt3k @kvcrawford is there anyone who could quickly try it?

kt3k:
@rahul-winner 
We have 2 doc sites: one for c3js.org (which requires ruby) and one for development (which only requires node).

http://localhost:8080/samples/zoom_category.html belongs to the latter and it can be invoked with the command `npm start`.

BTW after applying the patch, the example seems working on my machine. @panthony Could you describe how it breaks it?

panthony:
@kt3k It become unresponsive to mouse events, can't zoom or move around.

If it works for you it may be something else on my part but I can't re-test for now :-/

kvcrawford:
It's working for me on my branch in the PR.

@panthony you might need to build it first: `npm run dist && npm run start`

rahul-winner:
@kvcrawford @panthony @kt3k 
Have we concluded it? For me it looks like a good fix.

panthony:
I confirm the fix is actually working for me and that the zoom_category properly work.

Sorry for the lost time.

kt3k:
@panthony Thank you for confirming it again!

Now we seem on the same page 👍 Thank you all!

PR #2389, opened 2018-06-11
Refactors circular dependencies

amenadiel:
Curently, building with rollup using `npm run build:js` shows two warnings in the output 

> (!) Circular dependency: src/core.js -> src/axis.js -> src/core.js
> (!) Circular dependency: src/core.js -> src/axis.js -> src/util.js -> src/core.js

I've refactored the code in the following way:
 - `src/util.js` contains the root `c3` object, as well as modules `Chart` and `Component` so it doesn't need to import from `src/core`
 - `src/chart.js` imports everything it needs from `./class` and `./util` so it doesn't import from `./core` thus removing the circular dependency
 - `src/core.js` imports also `c3`, `Chart` and `Component` from  `src/util.js`
 - Added `src/chart-internal.js` which in turn imports `d3`. Every other module doesn't use `d3` directly so this is the only component that should access `d3`.
 - I've changed the way in which `d3` is accessed, since using it as

```js
$$.d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;  
```
isn't the proper way to deal with external modules in the ES6 approach. Instead, it should be imported and listed in `rollup.config.js` as global/external

 - Also, in `rollup.config.js` I added the `file` property to the output object, because [it is the reccomended way to point to the generated file](https://rollupjs.org/guide/en#using-config-files), so you can just call `rollup -c`  instead of doing `rollup -c > c3.js` (this approach also allows to output to different files, such as an UMD build along an ES6 build, if you ever wanted to do so). Besides, this results in a friendlier build output avoiding to have the whole source code piped to STDOUT. The `package.json` has been modified accordingly.

### PD

The resulting UMD module (not included in the commit as per the PR directives) will either use global d3 or require it as:

```js
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory(require('d3')) :
    typeof define === 'function' && define.amd ? define(['d3'], factory) :
    (global.c3 = factory(global.d3));
}(this, (function (d3) {
   ...
})));
```

Which accomplishes the same purpose of the old approach

```js
$$.d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;  
```

This also allows an explicit AMD dependency using `define` instead of `require`, although requirejs and SystemJS should support using CommonJS requires as rell.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,9 +4,13 @@ import pkg from './package.json'
 export default {
     input: 'src/index.js',
     output: {
+        file: 'c3.js',
         name: 'c3',
         format: 'umd',
-        banner: `/* @license C3.js v${pkg.version} | (c) C3 Team and other contributors | http://c3js.org/ */`
+        banner: `/* @license C3.js v${pkg.version} | (c) C3 Team and other contributors | http://c3js.org/ */`,
+        globals:{
+            d3:'d3'
+        }
     },
     plugins: [babel({
         presets: [['es2015', {
@@ -15,5 +19,6 @@ export default {
         plugins: [
             'external-helpers'
         ]
-    })]
+    })],
+    external: ['d3'],
 };
--- a/src/axis.js
+++ b/src/axis.js
@@ -1,6 +1,5 @@
 import CLASS from './class';
-import { Component } from './core';
-import { isValue, isFunction, isString, isEmpty } from './util';
+import {isValue, isFunction, isString, isEmpty, Component } from './util';
 
 export var c3_axis_fn;
 export var c3_axis_internal_fn;
--- a/src/util.js
+++ b/src/util.js
@@ -1,4 +1,41 @@
-import { c3_chart_internal_fn } from './core';
+import {ChartInternal} from './chart-internal';
+
+export var c3 = { version: "0.6.2" ,
+    chart: {
+        fn: Chart.prototype,
+        internal: {
+            fn: ChartInternal.prototype,
+        }
+    }
+};
+
+export function Chart(config) {
+    var $$ = this.internal = new ChartInternal(this);
+    $$.loadConfig(config);
+
+    $$.beforeInit(config);
+    $$.init();
+    $$.afterInit(config);
+
+    // bind "this" to nested API
+    (function bindThis(fn, target, argThis) {
+        Object.keys(fn).forEach(function (key) {
+            target[key] = fn[key].bind(argThis);
+            if (Object.keys(fn[key]).length > 0) {
+                bindThis(fn[key], target[key], argThis);
+            }
+        });
+    })(c3.chart.fn, this, this);
+}
+
+c3.generate = function (config) {
+    return new Chart(config);
+};
+
+export function Component(owner, componentKey, fn) {
+    this.owner = owner;
+    c3.chart.internal[componentKey] = fn;
+}
 
 export var isValue = function (v) {
     return v || v === 0;
@@ -13,7 +50,7 @@ export var isString = function (o) {
     return typeof o === 'string';
 };
 export var isUndefined = function (v) {
-return typeof v === 'undefined';
+    return typeof v === 'undefined';
 };
 export var isDefined = function (v) {
     return typeof v !== 'undefined';
@@ -31,7 +68,7 @@ export var isEmpty = function (o) {
     return typeof o === 'undefined' || o === null || (isString(o) && o.length === 0) || (typeof o === 'object' && Object.keys(o).length === 0);
 };
 export var notEmpty = function (o) {
-    return !c3_chart_internal_fn.isEmpty(o);
+    return !c3.chart.internal.fn.isEmpty(o);
 };
 export var getOption = function (options, key, defaultValue) {
     return isDefined(options[key]) ? options[key] : defaultValue;
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "docs": "bundle exec middleman",
     "build": "npm run build:js && npm run build:css",
     "build:js": "npm run build:js:rollup && npm run build:js:uglify",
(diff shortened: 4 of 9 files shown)


src/util.js
```diff
@@ -31,7 +68,7 @@ export var isEmpty = function (o) {
     return typeof o === 'undefined' || o === null || (isString(o) && o.length === 0) || (typeof o === 'object' && Object.keys(o).length === 0);
 };
 export var notEmpty = function (o) {
-    return !c3_chart_internal_fn.isEmpty(o);
+    return !c3.chart.internal.fn.isEmpty(o);
```
amenadiel: You already have `c3.chart.internal.fn` defined in this module, so you don't need to import it's alias from `./core`

src/util.js
```diff
@@ -13,7 +50,7 @@ export var isString = function (o) {
     return typeof o === 'string';
 };
 export var isUndefined = function (v) {
-return typeof v === 'undefined';
+    return typeof v === 'undefined';
```
amenadiel: This line was indented incorrectly

src/axis.js
```diff
@@ -1,6 +1,5 @@
 import CLASS from './class';
-import { Component } from './core';
-import { isValue, isFunction, isString, isEmpty } from './util';
+import {isValue, isFunction, isString, isEmpty, Component } from './util';
```
amenadiel: Component is exported by `./util` so you don't need to import it from `./core`  anymore

rollup.config.js
```diff
@@ -4,9 +4,13 @@ import pkg from './package.json'
 export default {
     input: 'src/index.js',
     output: {
+        file: 'c3.js',
         name: 'c3',
         format: 'umd',
-        banner: `/* @license C3.js v${pkg.version} | (c) C3 Team and other contributors | http://c3js.org/ */`
+        banner: `/* @license C3.js v${pkg.version} | (c) C3 Team and other contributors | http://c3js.org/ */`,
+        globals:{
+            d3:'d3'
```
amenadiel: This will list `d3` as a dependency in the top of the UMD build, instead of requiring `d3` from `ChartInternal`

kt3k:
Thanks for the contribution! I agree with most changes.

But I'd like to keep the timing of requiring d3 unchanged. Some users seem loading c3 and d3 in parallel, and this change breaks that situation. (We did the same change before and that caused the problem. See https://github.com/c3js/c3/issues/2060 ) So I'd prefer keep it requiring inside ChartInternal constructor.

Other part seems good to me :+1:

amenadiel:
I see. So the edge case scenario would be if someone loaded both d3 and c3 using `async defer`

But in that case, if you had no control over which script loaded first, you wouldn't either know when C3 has finished loading, and therefore you couldn't instance a new chart, unless you rely on setting a timeout.

I don't know... keeping a require among ES6 modules seems suboptimal. I can rollback that specific part for this PR to be merged, anyway

kt3k:
> you wouldn't either know when C3 has finished loading, and therefore you couldn't instance a new chart, unless you rely on setting a timeout.

You are right. I guess probably they instanciate the chart after `DOMContentLoaded` or `$(document).ready()` (I think that's a common manner in frontend dev).

> I don't know... keeping a require among ES6 modules seems suboptimal. I can rollback that specific part for this PR to be merged, anyway

I would appreciate if you rollback that part 🙏

amenadiel:
Okay, I restored the way `d3` is required to

```js
$$.d3 = window.d3 ? window.d3 : typeof require !== 'undefined' ? require("d3") : undefined;  
```

However, regarding the inclusion of `d3`, I'll be honest: 

I was working in a branch that bundled c3 along d3. The wheight of the unminified build was roughly 900 kb. (which is the sum of the unminified builds of c3 and d3, give or take). **BUT**, if you only import the D3 components you really use:

```js
export var d3 = {
     arc,
     area,
     brushSelection: brushSelection,
     brushX: brushX,
     brushY: brushY,
     csv: csv,
     curveBasis,
     curveBasisClosed,
     curveBasisOpen,
     curveBundle,
     curveCardinal,
     curveCardinalClosed,
     curveCardinalOpen,
     curveLinear,
     curveLinearClosed,
     curveMonotoneX,
     curveStep,
     curveStepAfter,
     curveStepBefore,
     customEvent: customEvent,
     drag: drag,
     easeLinear: easeLinear,
     extent: extent,
     get event () { return event; },
     get format () { return format; },
     get formatPrefix () { return formatPrefix; },
     get timeFormat () { return timeFormat; },
     get timeParse () { return timeParse; },
     get utcFormat () { return utcFormat; },
     get utcParse () { return utcParse; },
     interpolate: interpolate,
     json: json,
     line,
     max: max,
     merge: merge,
     min: min,
     mouse: mouse,
     pie,
     rgb: rgb,
     scaleLinear: scaleLinear,
     scaleTime: scaleTime,
     schemeCategory10: schemeCategory10,
     select: select,
     selectAll: selectAll,
     selection: selection,
     selector: selector,
     selectorAll: selectorAll,
     set: set,
     timeSecond: timeSecond,
     transition: transition,
     tsv: tsv,
     zoom: zoom,
     zoomIdentity: zoomIdentity,

 };
```

The bundle size is ~ 650 kb, so you shave off 250kb from d3. It `could` be useful if someone doesn't want to require `d3` via a script tag. You could have extra `c3.bundle.js` and `c3.bundle.min.js` scripts available for that particular use case. 

You can see those changes at https://github.com/HuasoFoundries/c3/commit/0f3e4d0118b57456da77157c677e52432fadc80d

kt3k:
Hmm..smaller bundle looks interesting. It's good for lower bandwidth environment and possibly some users might be interested using it...
But, on the other hand, I feel slimmed version of d3 is a bit hard for us to keep maintaining in the future...

Anyway we take this refactoring first. Thanks for the contributions!

Issue #2390, opened 2018-06-12
Gauge is not showing completely in half circle, the background arc goes missing. The gray color towards the end goes missing.


rahul-winner:
* **C3 version**:  0.6.1
* **D3 version**: 5.4.0
* **Browser**: Chrome 66.0.3359.117 (Official Build) (64-bit)
* **OS**: Windows

I noticed after updating to latest version of C3 and D3. The gray color from background area on Gauge is not appearing.

Please check this jsfiddle
http://jsfiddle.net/rahul_winner/r9Lfkamz/

**Before Update**
![image](https://user-images.githubusercontent.com/4982751/41413074-9df67786-6fff-11e8-81d6-917f548a3352.png)

**After Update**
![image](https://user-images.githubusercontent.com/4982751/41413191-f1ec2cdc-6fff-11e8-8bad-61b8cb10e018.png)

Could you please take a look at it and address it. It will be a great help.

rahul-winner:
does someone have any clue about it?

1Jesper1:
I would also like to know this. Seems the path is empty under class _c3-chart-arcs-background_

rahul-winner:
@1Jesper1 
Exactly, I noticed that. Do you have any solution for this? 

Kindly let me know if you are asking for some other information.

1Jesper1:
No, I noticed the gauge is showing completely is C3 version 0.4.11.

rahul-winner:
It is working till the time D3 is of 3.x where C3JS has 0.4.x version. 

As we update C3JS to latest which is using D3 5.x, we get to see this issue. I looked into the code however am not able to fix it as I'm not much familiar with D3.

rahul-winner:
created PR https://github.com/c3js/c3/pull/2401 for fixing this issue. Could anyone please merge it to master?

kt3k:
Released the fix #2401 as [0.6.3](https://github.com/c3js/c3/releases/tag/v0.6.3). Thank you @rahul-winner !

Issue #2343, opened 2018-04-12
arc background not drawn correctly


EnricoVogt:
C3 version: 0.5.3

At the current version it seems the background for an arc is not drawn.

Here an example with the current version:
https://codepen.io/anon/pen/LdvrJW

And here an example with the old "working" one:
C3 version: 0.4.21

https://codepen.io/anon/pen/QmPxzP

caique0202:
Same here. Any update?

kt3k:
Released the fix #2401 as [0.6.3](https://github.com/c3js/c3/releases/tag/v0.6.3). Thanks for reporting!

PR #2401, opened 2018-07-05
The path for background arc needs to be supplied with data to get it rendered. This went missing.

rahul-winner:
This is for fixing the issue https://github.com/c3js/c3/issues/2390

Solution : The path for background arc needs to be supplied with data to get it rendered.

### DIFF
--- a/src/arc.js
+++ b/src/arc.js
@@ -471,8 +471,7 @@ c3_chart_internal_fn.redrawArc = function (duration, durationForExit, withTransf
     if (hasGaugeType) {
         var index = 0;
         backgroundArc = $$.arcs.select('g.' + CLASS.chartArcsBackground).selectAll('path.' + CLASS.chartArcsBackground).data($$.data.targets);
-        backgroundArc.enter().append("path");
-        backgroundArc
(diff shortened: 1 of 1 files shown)


rahul-winner:
@kt3k 
did we take any decision for merging it to master? I believe you are the correct person who can provide some information. 
Pls let me know if you are seeing any functional deviation or issue here. 

Thanks a lot !

kt3k:
@rahul-winner 
I confirmed this change fixes the example in #2390 (and in #2343). Thank you for the contribution!

Do you think we can add any test case to prevent this regression?

rahul-winner:
@kt3k 
I'll explore if we could have test case for ensuring this behavior going further and will create PR. 

Thanks a lot for expediting this fix !

rahul-winner:
@1Jesper1 thanks for your comment.

Apparently some more work be required on background arc after moving on to recent D3. I'm investigating it and will try to fix it. Please feel free to contribute if you are seeing any fix.

rahul-winner:
let's track the resize issue here https://github.com/c3js/c3/issues/2409

PR #2416, opened 2018-07-20
Improved scrollToHash function code with specific selector on reference page of official website

adityasharma7:
When there is a link to content in column-content, the function may scroll to the wrong location. Check #legend-hide

https://c3js.org/reference.html#legend-hide


kt3k:
Thanks for catching this! Very helpful!

kt3k:
This will be fixed at the next release (because the site updates only when the new tag is created.)

kt3k:
This is deployed to c3js.org. [#legend-hide](https://c3js.org/reference.html#legend-hide) seems navigating to the right place! 👍

PR #2423, opened 2018-07-23
Change location of temporary build files

### LINKED ISSUE #2076: Move built files to dist, remove from version control

kt3k:
This PR moves the temporary built files under `tmp/` directory and makes symlinks point the new files.

The main dist files (`{c3,c3.min}.{css,js}`) only appears in diff when you build with `dist` command. So we can avoid unintentional update of dist files in PRs.

This doesn't change the location of main dist files. The fiddles pointint those files don't need to be updated.

This also introduces `npm-run-all` for organizing run scripts better.

cc @panthony 

---
closes #2076


kt3k:
closed in favor of #2424

PR #2428, opened 2018-07-24
Resolved sidemenu-content mismatch on reference page of official website

adityasharma7:
When browsing on reference page https://c3js.org/reference.html , there are some references in side menu for which there is no content, for some others content was present but no reference in side menu & some references were mispositioned
Added references for data.labels.format, data.ondragstart, data.ondragend
Updated axis.y.type position as it was mispositioned
Commented api.category, api.categories, api.tooltip.show, and api.tooltip.hide as no description content found


kt3k:
@adityasharma7 
Thanks for fixing this!

LGTM

Issue #2426, opened 2018-07-24
Separate publishing docs and test ci


kt3k:
- We want to allow PR's build to use repo:status token for sending back the preview URL to PR's status.
- If we allow it on circle CI, the deploy key for github pages is also given to PR's build (which is not an acceptable option)
- To avoid the above, we need to separate the roles of CI to travis and circleci
  - circle ci for PR build and travis for gh-pages deployment

---
tasks
- [x] deploy from travis
- [x] remove deploy settings and deploy credentials in circle
- [x] allow access to secrets in circle
- [x] send status back from circle to github

kt3k:
The last part is solved by #2432.

PR #2430, opened 2018-07-24
Store artifact of docs

kt3k:
This PR is the second part of #2426.

This PR
- stops publishing the doc site from circle ci (now which is being published from travis ci)
- instead stores the doc site as artifact on circle ci


kt3k:
Seems working.

Note: doc site on circle doesn't work for now because the path reference inside the site is root relative like `/css/c3.css`.

PR #2432, opened 2018-07-24
Send status back from circle ci

kt3k:
This PR sends back the preview site's URL as a commit status from circle ci.


kt3k:
`circleci/htdocs` status seems working as intended.

PR #1500, opened 2015-12-10
Added support to zoom types (scroll and drag) with option to disable default zoom behavior

mistic:
New api for zoom:

``` javascript
zoom: {
    enabled: true,
    type: 'drag', // can be [drag, scroll]
    disableDefaultBehavior: false // if yes chart don't zoom but user will have x domain of the zoom box. This is more util for drag zoom type.
}
```

PR #2439, opened 2018-08-03
Add zoom type 'drag' option (based on #1500)

kt3k:
This PR is a continuation of #1500. #1500 is not mergeable because of d3's breaking changes and c3's refactoring. I did the following to fit the feature to the current code base.

- Rebased #1500 
- Updated d3 API calls to follow the breaking changes between v3 and v5 (and did some refactoring)
- removed disableDefaultBehavior config because I couldn't find the use case of `zoom.disableDefaultBahavior: true`

The preview of this change is available [here (circleci artifact)](https://468-11496279-gh.circle-artifacts.com/0/htdocs/samples/zoom_type.html)

@mistic 
Could you check if the above page's behavior reflects your original idea of the feature?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,10 @@ components
 
 # coverage report
 /coverage
+
+# OS related
+.DS_Store
+
+# IDE related
+.idea
+
--- a/htdocs/index.html
+++ b/htdocs/index.html
@@ -374,6 +374,9 @@
               <a href="./samples/zoom_type.html">
                 Change Zoom Type
               </a>
+              <a href="./samples/zoom_type_disable_default_behavior.html">
+                Zoom Types with disableDefaultBehavior: true
+              </a>
               <a href="./samples/zoom_category.html">
                 Zoom on category axis
               </a>
--- a/htdocs/index.html
+++ b/htdocs/index.html
@@ -371,6 +371,9 @@
               <a href="./samples/zoom.html">
                 Enable zoom
               </a>
+              <a href="./samples/zoom_type.html">
+                Change Zoom Type
+              </a>
               <a href="./samples/zoom_category.html">
                 Zoom on category axis
               </a>
--- a/htdocs/samples/zoom_type.html
+++ b/htdocs/samples/zoom_type.html
@@ -1,0 +1,62 @@
+<html>
+  <head>
+    <link rel="stylesheet" type="text/css" href="/css/c3.css">
+  </head>
+  <body>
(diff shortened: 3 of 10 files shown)


mistic:
@kt3k the code looks okay and the features are there. I think we just should include the `disableDefaultBehavior` behaviour again (you can think on another name if you found this one a bit difficult to grasp the real meaning). I found this feature very useful for the case where we have aggregated data, and so the drag zoom will only return the x-domain, we re-quest the new data and then reload the chart with the new data.

kt3k:
@mistic I restored `zoom.disableDefaultBehavior` option. [Here](https://491-11496279-gh.circle-artifacts.com/0/htdocs/samples/zoom_type_disable_default_behavior.html) is the example of `disableDefaultBehavior: true` on both `scroll` and `drag` types.

mistic:
@kt3k perfect! I think it's ready to merge whenever you want :)

kt3k:
@mistic Thanks for confirming! 🙏 
I appreciate your patience. Thank you again for the original idea and initial implementation!

mistic:
No problem @kt3k :) I believe this will add value to c3js! Thanks for the effort on this.

mageshveeraa:
@kt3k @mistic  If I set zoom type as 'drag', tooltip not showing.. 

And, another problem I am facing is: Every time I zoom the chart zoom out to orginal state and zooming... I need multi level zoom... for example... First I zoom 1 to 2pm.. then I want to 1:45 to 1:55pm...

EliasC10:
> @kt3k @mistic If I set zoom type as 'drag', tooltip not showing..
 
@mageshveeraa i have the same problem, did you find a solution for that?

Federico-G:
* **C3 version**: v0.6.7
* **D3 version**: v5.7.0
* **Browser**: Chrome 69 & Firefox 62
* **OS**: Ubuntu 16.04 (updated)

@mageshveeraa I confirm the bug, and I think that multi level zoom should be the behavior

@kt3k @mistic I also found that zoom.enable is not working with this type of zoom

You can try it in https://c3js.org/samples/interaction_zoom_by_drag.html
In the console, paste `chart.zoom.enable( false );`. It doesn't do anything

Thanks

mageshveeraa:
@EliasC10  No, I didn't find the solution yet. Now I am using zoom by scroll only.

@Federico-G  Yes, I mean multi-level zoom behavior is not working.

Issue #2362, opened 2018-05-03
Bar chart tooltip is visible only when bar hovered


ilearnjs:
Hi all.  

A have a problem with bar chart tooltips.  

For the old versions of c3(0.4.21), d3(3.5.17) tooltip is visible if cursor is above or below bar rectangle  
![screenshot_12](https://user-images.githubusercontent.com/20753143/39577057-ef24a9d0-4ee8-11e8-86cc-0281d341731a.png)  
https://jsfiddle.net/r59fmaa0/

For the new versions c3(0.5.3), d3(4.12.0) tooltip is not visible in the same case  
![screenshot_11](https://user-images.githubusercontent.com/20753143/39578057-caf081e4-4eeb-11e8-8ca9-754abe5ee1d7.png)
https://jsfiddle.net/wx7xr9u3/

C3 version: 0.5.3
D3 version: 4.12.0
Browser: Google Chrome Version 65.0.3325.181 (Official Build) (64-bit)
OS: Windows 10

Thanks for support

Sazarov:
Has anyone found a workaround for this? I've been using c3 0.4 for a while but I'll have to upgrade at some point. Didn't find anything on whether this is an intentional change or not either. It seems to be unintentional because this is definitely worsening the user experience.

fuwaneko:
Any news on this? It blocks us from upgrading to newer version of D3 and C3.

MarkusLaut:
We also need this.

Jupiterrr:
I ran into #2391 which @panthony markted as a duplicate of this issue. Its seems to me that these two issues are both related to `findClosest()`
https://github.com/c3js/c3/blob/9556e38b41402c661c8e6d5c608d755ac2da84d5/src/data.js#L328-L355
which is called by https://github.com/c3js/c3/blob/9556e38b41402c661c8e6d5c608d755ac2da84d5/src/interaction.js#L64
So `findClosest()` is supposed to return the closes points or bars to the cursor. For bar charts only the actual hovered bar is returned and for line charts the closes point in a radius of 10 pixels is returned. Which in both cases might not be the desired behaviour.

1. line charts
I think a better behaviour is to ignore the y-axis and only uns the x distance of points and cursor.  Also consider ignoring `$$.config.point_sensitivity`  for this usecase, so `findClosest()` always returns the closest point on the x-axis and not only the closest within a minimum distance.

2. bar charts
[...]

PR #2476, opened 2018-09-13
Display tooltip nearest cursor on non-bar chart

### LINKED ISSUE #2391: Hover over entire chart to show tooltip
* **C3 version**: 0.5.4
* **D3 version**: 5.2.0
* **Browser**: FF
* **OS**: MacOS

I think the [Google Group is dead](https://groups.google.com/forum/#!topic/c3js/VC2hKY44h1Q). Right?

I have an [older c3 instance][1] where you can hover over the chart and the tooltip is showing no matter where you point at on the chart.

In a [newer instance][2] this behaviour is gone.

Now the tooltip is only shown when hitting the datapoint excately.

Is there a option for this?

  [1]: https://aare.guru/aare-statistiken
  [2]: http://c3js.org/samples/timeseries.html

ajb32x:
I want to resolve ~~2362~~ 2391, this displays the tooltip for the closest point by horizontal distance to the cursor.
fixes #2391

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -351,7 +351,7 @@ ChartInternal.prototype.findClosest = function (values, pos) {
     // find closest point from non-bar
     values.filter(function (v) {
         return v && !$$.isBarType(v.id);
-    }).forEach(function (v) {            
+    }).forEach(function (v) {
         var d = $$.horizontalDistance(v, pos);
         if (d < minDist || !minDist) {
             minDist = d;
--- a/src/data.js
+++ b/src/data.js
@@ -335,7 +335,7 @@ ChartInternal.prototype.horizontalDistance = function(data, pos) {
 };
 ChartInternal.prototype.findClosest = function (values, pos) {
     var $$ = this,
-        minDist,
+        minDist = Infinity,
         closest;
 
     // find mouseovering bar
@@ -353,7 +353,7 @@ ChartInternal.prototype.findClosest = function (values, pos) {
         return v && !$$.isBarType(v.id);
     }).forEach(function (v) {
         var d = $$.horizontalDistance(v, pos);
-        if (d < minDist || !minDist) {
+        if (d < minDist) {
             minDist = d;
             closest = v;
         }
--- a/src/data.js
+++ b/src/data.js
@@ -325,9 +325,17 @@ ChartInternal.prototype.findClosestFromTargets = function (targets, pos) {
     // decide closest point and return
     return $$.findClosest(candidates, pos);
 };
+ChartInternal.prototype.horizontalDistance = function(data, pos) {
+    var $$ = this,
+        config = $$.config,
+        xIndex = config.axis_rotated ? 1 : 0,
+        x = $$.x(data.x);
+
+        return Math.abs(x - pos[xIndex]);
+};
 ChartInternal.prototype.findClosest = function (values, pos) {
     var $$ = this,
-        minDist = $$.config.point_sensitivity,
+        minDist,
         closest;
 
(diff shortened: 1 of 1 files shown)


kt3k:
The preview example charts are available [here](https://540-11496279-gh.circle-artifacts.com/0/htdocs/index.html), but it doesn't seem fixing the problem reported in #2362.

src/data.js
```diff
@@ -325,9 +325,17 @@ ChartInternal.prototype.findClosestFromTargets = function (targets, pos) {
...
+        config = $$.config,
+        xIndex = config.axis_rotated ? 1 : 0,
+        x = $$.x(data.x);
+
+        return Math.abs(x - pos[xIndex]);
+};
 ChartInternal.prototype.findClosest = function (values, pos) {
     var $$ = this,
-        minDist = $$.config.point_sensitivity,
+        minDist,
```
danidin: Seems like `minDist` is never defined

src/data.js
```diff
@@ -344,13 +352,12 @@ ChartInternal.prototype.findClosest = function (values, pos) {
     values.filter(function (v) {
         return v && !$$.isBarType(v.id);
     }).forEach(function (v) {
-        var d = $$.dist(v, pos);
-        if (d < minDist) {
+        var d = $$.horizontalDistance(v, pos);
+        if (d < minDist || !minDist) {
```
Jupiterrr: You could initialize minDist with `Infinity` instead.

ifokeev:
merge please

9teen90nine:
need this merged too!

arnage:
great! merge it please

adamkasprowicz:
:D btw if you can't wait for the merge - change this part of the code manually in node_modules. Works fine 👍 *unless you're building your bundle in CI pipeline

ifokeev:
I found that PR doesn't resolve tooltip showing correctly. 
I noticed point_sensitivity config variable that isn't documented. 
You have to set 
```
        point: {
          sensitivity: 150
        },
```

and it will be actually the same behaviour as in the previous versions

ajb32x:
> I found that PR doesn't resolve tooltip showing correctly.
> I noticed point_sensitivity config variable that isn't documented.
> You have to set
> 
> ```
>         point: {
>           sensitivity: 150
>         },
> ```
> and it will be actually the same behaviour as in the previous versions

This PR should fix the issue in #2391 , which looks to have been mistakenly marked as a duplicate of #2362 . I'm happy to make changes, I would like issue #2391 resolved.

My changes actually remove the use of the point_sensitivity variable for non bar charts. I don't believe it is necessary to fix #2391 .

vlarkovich:
> The preview example charts (which includes this branch's change) are available [here](https://540-11496279-gh.circle-artifacts.com/0/htdocs/index.html), but it doesn't seem fixing the problem reported in #2362.

It seems that this PR will break the functionality of the tooltip when it is displayed separately for each point.
Try to open [not grouped tooltip](https://540-11496279-gh.circle-artifacts.com/0/htdocs/samples/tooltip_grouped.html), [scatter](https://540-11496279-gh.circle-artifacts.com/0/htdocs/samples/chart_scatter.html).

kt3k:
Maybe we should have both `findClosest` and `findClosestHorizontally` and use one of them depending on the options specified.

hln:
@vlarkovich is correct. When tooltip.grouped is set to false, the tooltip no longer works without having to directly hover over the points.

kt3k:
Closing in favor of #2582.

@ajb32x Thank you for tackling this.

PR #2495, opened 2018-10-01
Redraw on window focus

### LINKED ISSUE #2458: Data doesn't appear on an invisible tab
Steps to reproduce:
1.  Open any page with c3 chart on new tab (for example [c3js.org](https://c3js.org/samples/chart_combination.html))
2.  Do not navigate to it untill the page will not be fully loaded
3.  See the chart with empty data
image
4.  Moreover, if you hover mouse over the chart you can see "Uncaught INDEX_SIZE_ERR" error

When the chart is bulding on not visible tab the data will not appear because there is such condition:
https://github.com/c3js/c3/blob/22afe8150bb5020ea92d5a0ede944507354de05f/src/core.js#L681-L727

The code from inner else should be executed even if tab isn't visible.

* **C3 version**: 0.6.7
* **D3 version**: 5.6.0
* **Browser**: Chrome, Firefox, IE 11, Edge - latest versions
* **OS**: win10 x64

kt3k:
This PR changes to redraw on focus event on window. This prevent empty chart rendering when the page is opened in background tabs.

---
closes #2458

### DIFF
--- a/spec/c3-helper.js
+++ b/spec/c3-helper.js
@@ -3,8 +3,6 @@ import c3 from '../src';
 window.c3 = c3;
 
 window.initDom = function () {
-    'use strict';
-
     var div = document.createElement('div');
     div.id = 'chart';
     div.style.width = '640px';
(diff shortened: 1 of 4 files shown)


jlhedric:
Fair warning, this fix doesn't work in firefox. You can repro on the c3 website using v63.0.3 and v64.0 (latest stable version). It works if you wrap the redraw around a timeout, which I see you've done, but it needs to be longer than 10ms. Not sure if it varies by application but I had to set it to 75 before it consistently worked.

liu-minjie:
suppose there are two charts in tab/view
this fix will make chart area become smaller when the chart container is not visible

PR #2496, opened 2018-10-02
index parameter in tooltip.format.title #2392

### LINKED ISSUE #2392: Feature request: index parameter in tooltip.format.title
In the current version of C3 (v0.6.2), the [tooltip.format.title][1] function only presents the parameter of the x-coordinate. This is not helpful if you want the title to relate to some other property of the underlying data. Having the data-point index as an additional parameter would be useful here.

Note that other functions under tooltip.format provide much more data such as id and index.

[1]: http://c3js.org/reference.html#tooltip-format-title

awgaan:
## Title

Tooltip title format function is passed index

## Description

Resolves #2392

## Commit message

The tooltip title format function now gets the x-value as well as the index of the active point.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -131,7 +131,7 @@ ChartInternal.prototype.getTooltipContent = function (d, defaultTitleFormat, def
         }
 
         if (!text) {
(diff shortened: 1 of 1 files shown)


panthony:
Duplicate with #2553

PR #2517, opened 2018-11-01
fix regions.remove

danidin:


kt3k:
Thanks! Confirmed this fixes `.regions.remove()`.

ref: https://638-11496279-gh.circle-artifacts.com/0/htdocs/samples/data_region.html

ref: #2405

Issue #2467, opened 2018-08-29
Chart height determination by parent element not working


EmilijusS:
When chart height is unspecified it defaults to 320px, instead of parent element height.
It seems the reason is because ChartInternal.prototype.getParentHeight returns the height of chart itself. ChartInternal.prototype.getParentWidth on the other hand functions as intended.

https://github.com/c3js/c3/blob/947cd4794916109e33010e98c1dc0665be10c4e8/src/size.js#L75

mscraigloewen:
I put in a PR for this to fix this issue.

kt3k:
The fix has been released as [v0.6.11](https://github.com/c3js/c3/releases/tag/v0.6.11).

kt3k:
I think we can't change this behavior without breaking exsting apps. (See #2567 ) Please use a workaround like `chart.resize({ height: parentDom.clientHeight })` for now.

OlivierKamers:
Hi all,
I was wondering if there are plans for another attempt at a fix to avoid the need for a workaround?

PR #2541, opened 2018-12-03
Fixed chart height determination issue number 2467

mscraigloewen:
I changed the getParentHeight function to get the actual height of the parent element. I also removed the 'max-height' CSS style that was applied so that the chart could be resized properly when the browser window is resized. 

All changes passed the current tests and do not require any new tests.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -919,8 +919,6 @@ ChartInternal.prototype.updateSvgSize = function() {
     $$.svg.select('#' + $$.clipIdForSubchart).select('rect')
         .attr('width', $$.width)
         .attr('height', brush.size() ? brush.attr('height') : 0);
-    // MEMO: parent div's height will be bigger than svg when <!DOCTYPE html>
-    $$.selectChart.style('max-height', $$.currentHeight + "px");
 };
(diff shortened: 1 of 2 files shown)


kt3k:
@mscraigloewen 
Thanks for looking into this. The change seems ok to me. Can you add an example under /htdocs/ directory, which illustrates the case this PR fixes?

mscraigloewen:
I added a sample called resize.html that illustrates the case that the PR fixes. 
You just drag the screen around and can watch the chart fit to the size of the container its in! 👍

kt3k:
Released as [v0.6.11](https://github.com/c3js/c3/releases/tag/v0.6.11).

kt3k:
@mscraigloewen 
Sorry, we have to revert this change because it seems breaking the resizing behavior in many situations. See #2567 for details.

In addition to the above problem, I found that this patch doesn't work if `#chart` div doesn't have `height: 100%`.

I think we need to find a different approach to solve #2467

mscraigloewen:
No problem! I'll look into it if I have time.

PR #2582, opened 2019-03-11
Added option to only use x postion for tooltips

### LINKED ISSUE #2391: Hover over entire chart to show tooltip

sebzap:
I want to fix #2391, this is a variation of pull request #2476. Short summary: i implemented an option `{ tooltip: { horizontal: true } }` to enable the `horizontalDistance` function from #2476, however my implementation still utilizes `point_sensitivity` as minimum distance.

As mentioned [here](https://github.com/c3js/c3/pull/2476#issuecomment-430184110), horizontal distance does not make sense in all cases e.g. scatter plot. This i way i implemented it as configurable option.

Why not just use high `point_sensitivity`? (like mentioned [here](https://github.com/c3js/c3/pull/2476#issuecomment-429848571)) 
As soon as data points don't have much distance between them the y position has an high impact on the (relative) distance to the mouse pointer. Look at following two screenshots of an chart with high point sensitivity. Note that i moved the mouse only a few pixels and the tooltip jumps 3 data points to the right. I'd like to have behaviour where the tooltip is on the closest x-data-point (implemented by this PR)
image
image

### DIFF
--- a/htdocs/index.html
+++ b/htdocs/index.html
@@ -305,6 +305,9 @@
               <a href="./samples/tooltip_grouped.html">
                 Show tooltip as each data
               </a>
+              <a href="./samples/tooltip_horizontal.html">
+                Show tooltip based on horizontal mouse position
+              </a>
             </div>
           </div>
         </div>
--- a/htdocs/samples/tooltip_horizontal.html
+++ b/htdocs/samples/tooltip_horizontal.html
@@ -1,0 +1,27 @@
+<html>
+  <head>
+    <link rel="stylesheet" type="text/css" href="../css/c3.css">
+  </head>
+  <body>
+    <div id="chart"></div>
+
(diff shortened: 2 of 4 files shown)


kt3k:
I'm very sorry for the delayed response. I confirmed the example works perfectly. The rationale for making this behavior optional makes sense to me.

Issue #2558, opened 2019-01-08
Memory leak when switching tabs


NickJSpencer:
* **C3 version**: 0.6.12
* **D3 version**: 5.7.0
* **Browser**: Chrome Version 68.0.3440.106
* **OS**: macOS High Sierra Version 10.13.6

Every so often, when I switch tabs away from the one using the c3 library, the javascript memory will begin to increase.

<img width="486" alt="screen shot 2019-01-07 at 3 15 02 pm" src="https://user-images.githubusercontent.com/22459831/50798989-2876ac00-128f-11e9-9923-eb68cdbfefd3.png">

I have narrowed this down to the c3 library scheduling transitions while away from the tab. I can't imagine this is a problem in my codebase because I never specify transitions on any c3 objects. Below are some images that might be helpful!

Screenshot of where the memory leak occurs. Note the array of scheduled transitions is very large.
<img width="1189" alt="screen shot 2019-01-07 at 3 12 37 pm" src="https://user-images.githubusercontent.com/22459831/50798885-c027ca80-128e-11e9-80ea-de80e2e48dc8.png">

Screenshot of what one of the scheduled transition objects looks like:
<img width="335" alt="screen shot 2019-01-07 at 3 16 30 pm" src="https://user-images.githubusercontent.com/22459831/50802789-9d9cae00-129c-11e9-983c-824935e279d8.png">

Screenshot of the memory snapshot on chrome tools if the memory leak is happening (~3 mins tabbed out of the tab with the leak):
<img width="1241" alt="screen shot 2019-01-07 at 4 45 17 pm" src="https://user-images.githubusercontent.com/22459831/50802836-dfc5ef80-129c-11e9-99b0-fb24ba055f21.png">

Screenshot of the memory snapshot on chrome dev tools if the memory leak doesn't occur as a reference:
<img width="1241" alt="screen shot 2019-01-07 at 4 49 45 pm" src="https://user-images.githubusercontent.com/22459831/50802809-c15ff400-129c-11e9-96d6-2cc49d0a575a.png">

xdev-developer:
yeah, same with linux chrome.

pstromberg98:
Ran into this as well. Looking more into it, I found out why (in my case) the `schedules` dictionary endlessly builds while tab is inactive. This is my first time looking into this library so I may have a few details wrong or missing due to my lack of understanding the entire c3 project.

[...]

PR #2589, opened 2019-03-22
Fixed memory leak when blurring tab during transitions

pstromberg98:
Fix for #2558. Simply checks if tab is active during the generateWait interval. If it isn't active, then it won't attempt transitioning, but will continue once the tab becomes active again.

For a detailed explanation of the bug, click [here](https://github.com/c3js/c3/issues/2558#issuecomment-475429026).


xdev-developer:
cool, thanks!

kt3k:
@pstromberg98 
Thank you for looking into this!

I confirmed the leak behavior exists with the flow example ( https://c3js.org/samples/api_flow.html ) and this fixes the issue. (Observed task manager's memory amounts for a while). Awesome fix!

The code looks good, but we use `rollup` for building the `c3.js` file, and actual source code is in `src/core.js` in this case. Can you move the change to that file?

pstromberg98:
@kt3k 
Oops! Sorry about that. The fix has been moved to the correct spot.

Thanks for taking a look at this and the kind words! Excited to see this merged in 😊.

PR #2595, opened 2019-04-03
Stanford Diagrams

FreezinG117:
Add support for [Stanford Diagrams](https://gssc.esa.int/navipedia/index.php/The_Stanford_%E2%80%93_ESA_Integrity_Diagram:_Focusing_on_SBAS_Integrity).


kt3k:
Wow, this is an amazing patch 🤯

I opened `chart_stanford.html` and `chart_stanford_custom_elements.html`, and both show the error:
```
Error: number of elements necessary for Stanford graph type is invalid, must be: json {somekey, somekey, epochs}
```
Can you check these examples?

I will review soon in more details.

FreezinG117:
Ups, it should work now 🙂

spec/stanford-spec.js
```diff
@@ -0,0 +1,157 @@
...
+
+        it('should return 100% if the region has all the epochs', function () {
+            var region = [
+                {x: 0, y: 0},
+                {x: 60, y: 0},
+                {x: 60, y: 60},
+                {x: 0, y: 60}
+            ];
+
+            var result = chart.internal.countEpochsInRegion(region);
```
kt3k: Nice unit test 👍

src/type.js
```diff
@@ -55,7 +55,11 @@ ChartInternal.prototype.isBarType = function (d) {
 };
 ChartInternal.prototype.isScatterType = function (d) {
     var id = isString(d) ? d : d.id;
-    return this.config.data_types[id] === 'scatter';
+    return this.config.data_types[id] === 'scatter' || this.config.data_types[id] === 'stanford';
```
kt3k: I think this is confusing. How about creating another method `isScatterOrStanfordType` and replace the usages of `isScatterType` with it where necessary.

kt3k:
I commented one point. Please check.

This has nice tests, nice documentation and examples. The implementation is enough independent of other part of source code. I think it's reasonable to include this chart to c3.

Thank you for the great contribution!

src/type.js
FreezinG117: @kt3k I did what you suggested, see if the changes I made work for you. Thanks!

kt3k:
released as v0.7.0. Thank you for the huge contribution! https://c3js.org/samples/chart_stanford.html

PR #2601, opened 2019-04-09
Run the tests on Node.js version 10

serima:
Node 8 is now a maintenance LTS, and I think that from now on we should test Node 10, which is the active LTS, in the main.
For the time being we try to run the tests in parallel.

https://github.com/nodejs/Release#release-schedule


serima:
Thank you very much for quick reviews and merging!

PR #2631, opened 2019-06-13
Make sure innerRadius is not NaN value

arnsvel:
Error: attribute dx: Expected length, "NaNpx".
Replication:
Happens every time lost focus from/to browser tab when gauge chart is dynamically generated.

Investigation:
w = config.gauge_width is has assigned value 10
$$.innerRadiusRatio is -Infinity
$$.innerRadius is NaN

### DIFF
--- a/src/arc.js
+++ b/src/arc.js
@@ -492,7 +492,8 @@ ChartInternal.prototype.redrawArc = function (duration, durationForExit, withTra
         backgroundArc
             .exit()
             .remove();
-
+        
+        $$.innerRadius = $$.innerRadius || 0;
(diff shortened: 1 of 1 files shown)


src/arc.js
```diff
@@ -492,7 +492,8 @@ ChartInternal.prototype.redrawArc = function (duration, durationForExit, withTra
         backgroundArc
             .exit()
             .remove();
-
+        
+        $$.innerRadius = $$.innerRadius || 0;
```
kt3k: I think this is not a right place to do this. How about checking this in `updateRadius`?

kt3k (commented):
Thank you for looking into this! one comment.

PR #2670, opened 2019-08-07
Add normalized stack support

### LINKED ISSUE #142: Normalized Stacked Chart
Hi Masayuki,
I was looking for a way to implement stacked chart expressed as percentages, similar to this example: http://bl.ocks.org/mbostock/3886394
However I was overwhelmed by so many intermediate steps you are using to calculate Y offset and height of bars. I'm confused so far...
A quick and dirty way to do this is to normalize data before drawing chart and then set Y range to 0-100%. However this is not suitable as I need to present both percentages and original values in labels on chart and in tooltips.
The same normalization should be also applicable to area charts.
Please advice what would be the best way to go and I'll be glad to contribute.

panthony:
Can be configured using:

```js
{
  data: {
    stack: {
      normalize: true
    }
  }
}
```

Defaults to false.

Closes #142


panthony:
@kt3k Hello there, I was looking to cache some function returns and I noticed the `cache.js` file.

Look how it is used I found that this is mostly used by `load()` through `cacheIds` but:

1. There are no documentation whatsoever nor issues about this
2. Weird to have a public API passing "cache ids"

Do you know how/if this is used ? Would it be possible to drop the support for it ?

My intent would be to modify the cache interface to be used in more contexts internally.

panthony:
Big part of untested added code is to be used when `padAngle` option will be available (see #2669).

Can't add a test for it right now :/

kt3k:
@panthony Personally I think these cache APIs don't belong to "public API" and we can change (or replace) them as long as the existing charts work.

panthony:
@kt3k One of the usage of this cache was "recently" added for gauge, but I think the value could be retrieved another way:

> return $$.levelColor($$.cache[id].values[0].value);

Otherwise, the only usage is `load({ cacheIds: [] })`. But I don't really know how it was (is?) used.

src/api.data.js
```diff
@@ -9,12 +10,32 @@ Chart.prototype.data = function (targetIds) {
 Chart.prototype.data.shown = function (targetIds) {
     return this.internal.filterTargetsToShow(this.data(targetIds));
 };
-Chart.prototype.data.values = function (targetId) {
-    var targets, values = null;
+
+/**
+ * Get values of the data loaded in the chart.
+ *
+ * @param {String|Array} targetId This API returns the value of specified target.
```
kt3k: 👍

src/util.js
```diff
@@ -51,3 +54,6 @@ export var notEmpty = function(o) {
 export var sanitise = function(str) {
     return typeof str === 'string' ? str.replace(/</g, '&lt;').replace(/>/g, '&gt;') : str;
 };
+export var mergeArray = function(arr) {
```
kt3k: How about calling this function `flattenArray`? (I think that's more common name for this)

src/util.js
```diff
@@ -51,3 +54,6 @@ export var notEmpty = function(o) {
...
+    return (arr && arr.length ? arr.reduce((p, c) => p.concat(c)) : []);
```
kt3k: I think this can be implemented like the below, and maybe it's easier to understand.

```js
return Array.isArray(arr) ? [].concat(...arr) : [];
```

kt3k:
The example looks great! 👍 

スクリーンショット 2019-08-10 19 38 32

panthony:
@kt3k I added a new exemple of a combinaison chart using stack normalisation per axis basis:

Capture d’écran 2019-08-13 à 15 38 50

src/util.js
panthony: done !

src/util.js
panthony: done !

shaswa:
Hi!
Thank you so much for this feature! I'm trying to use normalized stacked bar chart in a project. It works as expected except for the tooltip. The tooltip still shows the original values, instead of normalized or percentage values. Do I need to set anything else to display the tooltip values in percentages?

PR #2671, opened 2019-08-08
only skip animation if !isTabVisible, not the whole rendering

### LINKED ISSUE #2533: isTabVisible prevents chart being rendered
When a chart finishes its rendering in a tab that is currently not visible, all plots of the charts is not rendered when the tab becomes visible again.
As a temporary fix i changed the **ChartInternal.prototype.isTabVisible** method to always return true.

* **C3 version**:
0.6.8
* **D3 version**:
5.7.0
* **Browser**:
Chrome(latest)/IE11
* **OS**:
Windows

panthony:
It fixes a regression introduced by f4a163f7128ed9218265ed9ce5163d758d604302 and closes #2533

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -696,51 +696,49 @@ ChartInternal.prototype.redraw = function(options, transitions) {
         });
     }
 
-    if ($$.isTabVisible()) { // Only use transition if tab visible. See #938.
-        if (duration) {
-            // transition should be derived from one transition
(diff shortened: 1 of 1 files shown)


panthony:
Hum... I just saw this commit: 19a54ec413c5ae37ec4079c4303e38824a6fd967 but it did not fixed the issue for me.

Issue #2303, opened 2018-02-27
Performance issues with c3js


panthony:
I'm currently struggling with the poor performance of c3js as others do and there is a lot of reasons for that.

- displaying a single line chart (from `simple.html`) calls  `getBoundingClientRect` 99 times which has a significant cost ( #1645 )

- displaying a single line chart will also append and remove from DOM a dummy (to mesure font size) a dozen times per chart (probably because we call updateSizes N times per display)

- it creates and resizes absolutely everything, although there is a single line chart it also create, append and resize the nodes for regions, arcs, titles, grid, the Y2 axis and the X axis for subchart (#1591)

- each axis is created up to 5 times at render time, meaning for our line chart with 2 real axes we create in reality (with uncached sizing since it's a new instance etc.) 20 axes because we call updateScales/updateSizes multiple times at render time:

  - initWithData -> redraw -> updateDimension -> updateScales
  - initWithData -> updateDimension -> updateScales
  - initWithData -> initLegend -> updateLegendWithDefaults -> updateLegend -> updateScales
  - initWithData -> initLegend -> updateLegendWithDefaults -> updateLegend -> updateSizes
  - initWithData -> updateScales 

Basically the browser is taking all its time to modify dom, force relayout, mesure values, modify dom, force relayout, etc.. hundreds of times. (#1067)

There was attempt to cache values with `withoutCompute` for some parts but more often than not this is not used because the boolean is not passed.

I started trying to opt-out from unused stuff (starting with axes) and I could reduce the 99 calls to `getBoundingClientRect` to 28 which feels much better on big dashboards but I'm not 100% familiar with the code stack so I can't be sure I'm not breaking some edge cases here and there.

Is there anyone with some knowledge of the code stack that would be interested in trying to improve the performance? Is it worth it to try and do thing nicely and create PRs to this repository?

panthony:
@conceptualspace I removed the `withoutRecompute` and made it the default behavior, the cache is manually cleared when data is (re)loaded (https://github.com/cogniteev/c3/commit/48c7f5bf5ff276169167390d64a64268cc261977).

Also, I opt-out a bunch of stuff when not used like subcharts (https://github.com/cogniteev/c3/commit/33b7a85e23ed265ecca28137e72a7782a5e353df) or y2Axis (https://github.com/cogniteev/c3/commit/800164bfa7b22f0128c068dd4f5fb67670f5e033) which save a massing amount of works.

[...]

PR #2674, opened 2019-08-09
Opt-out from subchart DOM if not enabled

### LINKED ISSUE #358: hide/show subchart ?
Hello my friend,

Is there a easy way for hide/show the subchart under the chart ?

thanks :)

### LINKED ISSUE #2112: Valid svg subchart error (+fix)
A little bug that doesn't affect the running of c3 but rather any manipulation of the svg if produces...

lines 874-876 of core.js are currently

https://github.com/c3js/c3/blob/master/src/core.js#L874

```
    $$.svg.select('#' + $$.clipIdForSubchart).select('rect')
        .attr('width', $$.width)
        .attr('height', brush.size() ? brush.attr('height') : 0);
```

I've got a copy where I needed to change it to the code below to make sure the value '0' rather than 'undefined' was assigned to the height attribute. Otherwise when I later exported the chart svg the w3c validation service and inkscape both kicked up a fuss.

```
    $$.svg.select('#' + $$.clipIdForSubchart).select('rect')
        .attr('width', $$.width)
        .attr('height', brush.size() ? (brush.attr('height') || 0) : 0);    // mjg changed this line
```

### LINKED ISSUE #2503: Rotated chart with hidden subchart still able to interact with
* **C3 version**: 0.6.8
* **D3 version**: 5.7.0
* **Browser**: Chrome
* **OS**: Windows

I have a rotated bar graph with long x-tick values. I appears there is a hidden subchart that user can still interact with and zoom in with, even though no subchart or zoom is enabled. See my codepen below:

https://codepen.io/u119102/pen/NOpBmK?editors=1010

If you hover mouse over the middle of the word "January", mouse pointer "+" will show up, pull it down and the graph will zoom in (similar to a subchart functionality)

This does seem trivial in web-page view, but in mobile view, it's in the left side of the page where user tends to scroll with their finger, it can be quite irritating, especially since you cannot zoom all the way back out, without seeing the subchart. 

image

Please let me know if any other information need or it has already been solved.

Thanks.

panthony:
This PR does several things:

- do not append DOM's nodes for subchart if not enabled (Closes #2503 and first step for #2303)
- correct the 'undefined' height (Closes #2112)
- Add show/hide API methods (Closes #358)

### DIFF
--- a/docs/reference.html.haml
+++ b/docs/reference.html.haml
@@ -261,6 +261,9 @@
           = partial :reference_menu_item, locals: { id: 'api.axis.range' }
           = partial :reference_menu_item, locals: { id: 'api.legend.show' }
           = partial :reference_menu_item, locals: { id: 'api.legend.hide' }
+          = partial :reference_menu_item, locals: { id: 'api.subchart.isShown' }
+          = partial :reference_menu_item, locals: { id: 'api.subchart.show' }
+          = partial :reference_menu_item, locals: { id: 'api.subchart.hide' }
           = partial :reference_menu_item, locals: { id: 'api.zoom' }
           = partial :reference_menu_item, locals: { id: 'api.unzoom' }
           = partial :reference_menu_item, locals: { id: 'api.zoom.enable' }
@@ -4898,7 +4901,6 @@
         %h3
           = partial :reference_item_link, locals: { id: 'api.legend.hide' }
         %p Hide legend for each target.
-        %h5 Arguments:
         %div.sourcecode
           %pre
             %code.html.javascript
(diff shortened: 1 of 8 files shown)


panthony:
@kt3k I just don't get it, I installed from scratch RVM, bundler, etc.. from the `0.7.3` tag and the documentation site served by `watch:docs` does not work.

I have a wonderful `NameError: uninitialized constant #<Class:0x00007fdfd8214c78>::EXPR_BEGEXPR_BEGEXPR_BEGEXPR_BEGEXPR_CMDARGEXPR_DOTEXPR_ARGEXPR_BEG` error located here:

```
%a( href="##{ id.gsub(/\./, '-') }")< #{ id.gsub(/(api|class)\./, '') }
```

I played around and the issue is with `%a()` .. maybe it misses `id` ? Result of a breaking change due to an unwanted update ?

I really have no clue I have 0 knowledge of the stack used to generate the doc.

kt3k:
@panthony Which ruby version do you use? You need ruby 2.4.4 for building the current document site (I think it can't be built with ruby 2.5 or above).

panthony:
@kt3k I'll check that on monday, I do not have my laptop with me this week-end but I'll likely have the most recent version of ruby.

Good to know I should not have a version too recent, thanks :)

Issue #2213, opened 2017-10-30
`attrTween` does not get removed properly on `destroy`


stephencookdev:
* **C3 version**: 0.4.18
* **D3 version**: 3.5.6
* **Browser**: all
* **OS**: all

[Here's a jsfiddle reproducing the issue](http://jsfiddle.net/stephencookdev/r6tsc8yt/).

You'll see it produce an error similar to `Uncaught TypeError: Cannot read property 'data_types' of null` (depending on your browser).

Specifically, [this](https://github.com/c3js/c3/blob/64a134cc4407d4ef76998f28ce455c5fd118766d/src/arc.js#L378) is what I've seen causing the issue.

It looks like the `attrTween` callback continues to be called after `chart.destroy()` is invoked, which then obviously breaks.

I'm happy to raise a fix for this, but would definitely appreciate if anyone could point me in the right direction in terms of APIs exposed in c3/d3 to deregister this tween

Jorrex:
I was experiencing the same issue where a chart was redrawn pretty fast due to input changing all the time.
This caused the same error and I applied this fix and can confirm that the error is not being thrown anymore.

PR #2681, opened 2019-08-16
Fix: tween callbacks run after chart is destroyed.

cythrawll:
#2213

- This causes some issues when unloading arc type charts

previous PR was abandoned, but we still see this error quite a bit, would like to see this cleaned up.


panthony:
@cythrawll Looks good! Thanks!

PR #2151, opened 2017-08-23
Track x mouseovers

nbdavies:
Fixes issues described in https://github.com/c3js/c3/issues/1745

Here's a jsfiddle showing the problem: http://jsfiddle.net/nbdavies/2gad17bL/
Try moving from one blue bar to another via the X axis.

Issue #2075, opened 2017-06-08
Stop using Deprecated .pathSegList and remove SVGPathSeg polyfill


kt3k:

kt3k:
ping @masayuki0812

Issue #1658, opened 2016-04-18
Stacked Bar Onclick captures multiple datapoints 


raphaelgera:
When capturing onclick event with stacked bar, if clicking on/near the border between two elements the click fires twice.

Use the docs and to reproduce add to data:
 onclick: function(id, element){
            console.log(id);
        }

http://c3js.org/samples/chart_bar_stacked.html
![stackedbars onclick grabs both points](https://cloud.githubusercontent.com/assets/12085354/14618896/dada473e-0582-11e6-98fc-dc0c0d883bd3.png)

aendrew:
Oddly I'm getting a TypeError when using the onclick handler with the grouped bar chart. Marking as bug.

martingraham:
The problem seems to be that C3 calculates if a mouse click is within a bar using a bit of boundingbox maths in chart.internal.isWithinBar, and adds in an offset value that catches 'near misses' to this box. It does this for each bar at a given x value in the chart so several can actually report a click if it happens near adjacent bars. If this offset is fudged to say -1 the problem goes away but then really small / narrow bars are very difficult to click instead.

chart.internal.isWithinBar = function (that) {
        var mouse = this.d3.mouse(that), box = that.getBoundingClientRect(),
            seg0 = that.pathSegList.getItem(0), seg1 = that.pathSegList.getItem(1),
            x = Math.min(seg0.x, seg1.x), y = Math.min(seg0.y, seg1.y),
            w = box.width, h = box.height, offset = 2, // offset causes the multiple onclick problem
            sx = x - offset, ex = x + w + offset, sy = y + h + offset, ey = y - offset;
        return sx < mouse[0] && mouse[0] < ex && ey < mouse[1] && mouse[1] < sy;
    };

node-monk:
Updating to the latest version of C3 and D3 seemed to have fixed this for me (at least better than the last version I was using).

panthony:
Should be fixed by #2684

PR #2684, opened 2019-08-19
Restore tooltip behavior when grouped

### LINKED ISSUE #1201: Grouped tooltip ignored for scatter
Hi,

It seems that grouped tooltip is ignored for the scatter plot type.

Example

```
var chart = c3.generate({
    data: {
        columns: [
            ['data1', 30, 200, 100, 400, 150, 250],
            ['data3', 500, 320, 210, 340, 215, 125]],
        type: 'scatter'
    },
    tooltip: {
        grouped: true
    }
});
```

### LINKED ISSUE #1745: mouseover doesn't fire if xSort is false and only single data point plotted
In a bar chart, if data.xSort is set to false and you are passing in only a single data column in, then the mouseover function isn't triggered when moving from bar to bar. Because xSort is set to false, c3 thinks it is dealing with multipleXs. `$$.isMultipleX()` returns true if xSort is false, so `$$.generateEventRectsForMultipleXs()` gets called. That function compares `closest.id !== $$.mouseover.id` yet since there is only one data column, the id values will always be the same. 
http://codepen.io/mrnosal/pen/gMmzMK
c3 v0.4.11

### LINKED ISSUE #2362: Bar chart tooltip is visible only when bar hovered

### LINKED ISSUE #2641: LineChart: Data-events do not trigger if tooltip.horizontal is true
* **C3 version**: 0.7.2
* **D3 version**: 5.9.7
* **Browser**: Firefox 68
* **OS**: Fedora 30

* **Sandbox**: https://codesandbox.io/embed/nice-shape-9w8sv?fontsize=14

Whenever `tooltip.horizontal` is enabled the `data.onclick`, `data.onmouseover`, etc. do not trigger.

`data.onclick` only triggers if it's a point on the first series in `data.columns`

## Expected results:

1. Click on a point from `data2`
1. Expect alert to pop-up.

## Actual results:

1. Click on point from `data2`
1. Nothing happens

panthony:
In previous version of `c3js` when `tooltip_grouped` was true, the simple act of hovering above a category would highlight/show the data point(s) of that particular category.

But it was broken when c3js started to use a single rect for the whole chart.

This PR aims to restore this important behavior.

It also includes the fix proposed in #2151.

Closes #2362
Closes #1745
Closes #2641
Closes #1201

It removes one of the usage of pathSegList (#2075)

Left to do:

- [x] Add tests for timeseries

### DIFF
--- a/spec/interaction-spec.js
+++ b/spec/interaction-spec.js
@@ -46,6 +46,9 @@ describe('c3 chart interaction', function () {
             });
 
             describe('mouseover', function () {
+                let mouseoutCounter = 0;
+                let mouseoverCounter = 0;
+
                 beforeAll(function () {
                     args = {
                         data: {
@@ -56,10 +59,10 @@ describe('c3 chart interaction', function () {
                             ],
                             type: 'bar',
                             onmouseout: function() {
-                                window.mouseoutCounter += 1;
+                                mouseoutCounter += 1;
                             },
                             onmouseover: function() {
-                                window.mouseoverCounter += 1;
+                                mouseoverCounter += 1;
                             }
                         },
                         axis: {
@@ -69,23 +72,23 @@ describe('c3 chart interaction', function () {
                 });
 
                 beforeEach(function() {
-                    window.mouseoverCounter = 0;
-                    window.mouseoutCounter = 0;
+                    mouseoverCounter = 0;
+                    mouseoutCounter = 0;
                 });
 
                 it('should be undefined when not within bar', function () {
                     moveMouseOut();
 
-                    expect(window.mouseoutCounter).toEqual(0);
-                    expect(window.mouseoverCounter).toEqual(0);
+                    expect(mouseoutCounter).toEqual(0);
+                    expect(mouseoverCounter).toEqual(0);
                     expect(chart.internal.mouseover).toBeUndefined();
                 });
 
                 it('should be data value when within bar', function () {
                     moveMouse(31, 280);
 
-                    expect(window.mouseoutCounter).toEqual(0);
-                    expect(window.mouseoverCounter).toEqual(1);
+                    expect(mouseoutCounter).toEqual(0);
+                    expect(mouseoverCounter).toEqual(1);
                     expect(chart.internal.mouseover).toEqual({
                         x: 0,
                         value: 30,
@@ -99,8 +102,8 @@ describe('c3 chart interaction', function () {
                     moveMouse(31, 280);
                     moveMouseOut();
 
-                    expect(window.mouseoutCounter).toEqual(1);
-                    expect(window.mouseoverCounter).toEqual(1);
+                    expect(mouseoutCounter).toEqual(1);
+                    expect(mouseoverCounter).toEqual(1);
                     expect(chart.internal.mouseover).toBeUndefined();
                 });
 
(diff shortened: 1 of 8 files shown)


panthony:
I added more tests but codecov did not re-run, not sure if possible to make it re-run manually.

panthony:
@kt3k I was thinking about fixing #1658 too by moving [data_onclick](https://github.com/c3js/c3/pull/2684/files#diff-548b926079146c87b0ba47c42f0d3ab1R164) outside the loop and simply calling it once with `closest` which is supposed to be the closest data point to the mouse.

What do you think?

spec/interaction-spec.js
```diff
@@ -34,6 +45,82 @@ describe('c3 chart interaction', function () {
...
+                            }
+                        },
+                        axis: {
+                            rotated: false
+                        }
+                    };
+                });
+
+                beforeEach(function() {
+                    window.mouseoverCounter = 0;
```
kt3k: Do these variables need to be global? How about declaring them with `let` under `describe('mouseover'` scope?

spec/interaction-spec.js
```diff
@@ -1,8 +1,19 @@
...
     var chart, args;
 
+    const moveMouseOut = () =>
+        setMouseEvent(chart, 'mouseout', 0, 0, d3.select('.c3-event-rect').node());
+
+    const moveMouse = (x = 0, y = 0) =>
+        setMouseEvent(chart, 'mousemove', x, y, d3.select('.c3-event-rect').node());
+
+    const clickMouse = (x = 0, y = 0) =>
+        setMouseEvent(chart, 'click', x, y, d3.select('.c3-event-rect').node());
```
kt3k: 👏

spec/interaction-spec.js
```diff
@@ -108,4 +195,711 @@ describe('c3 chart interaction', function () {
...
+                                enabled: true,
+                                r: 8
+                            }
+                        }
+                    }
+                };
+            });
+
+            beforeEach(function() {
+                window.clickedData = [];
```
kt3k: I prefer putting this variable in more local scope as well (like `describe('line chart'`) to reduce the possibility of affecting the other tests in future.

kt3k:
@panthony Thank you for tackling this. This was one of the most requested fixes in c3js! The examples seem working nicely on chrome and firefox 👍 Testing with mouse interaction looks so great!

I left a few comments.

> I was thinking about fixing #1658 too by moving data_onclick outside the loop and simply calling it once with closest which is supposed to be the closest data point to the mouse.

Sounds good to me.

IceCreamYou:
Is there a way to avoid this behavior? I have charts with large tooltips, and with this new behavior where tooltips appear when you're not hovering a point/bar, the tooltips get in the way. In particular they get in the way of drag-selections and sometimes even seeing the data point you're trying to hover. I know that I can set `tooltip.grouped` to `false`, and I can override `tooltip.contents` to still display grouped info in that case, but I want to keep the behavior that if a point and bar have the same index, hovering over the bar also gives the point hover styles (and vice versa). If possible, I would prefer not to have to revert to manually toggling classes based on what's hovered or where the mouse is.

Issue #2407, opened 2018-07-09
minimum d3js


laurentdebricon:
* **C3 version**:
* **D3 version**:
* **Browser**:
* **OS**:

ahorek:
these works for me, it will reduce d3 size by around 50%
d3-array.js
d3-brush.js
d3-collection.js
d3-color.js
d3-dispatch.js
d3-ease.js
d3-format.js
d3-interpolate.js
d3-path.js
d3-scale-chromatic.js
d3-scale.js
d3-selection.js
d3-shape.js
d3-time-format.js
d3-timer.js
d3-transition.js
d3-zoom.js

davidhund:
I am _very_ interested in a custom build of C3, importing only necessary D3 modules for our use-case.
(E.g. only time-series base line-charts with tooltips, zoom and selection)

However: I could not get it to work properly in a Rollup/ES6 imports project.

Any working examples/tips on how to build a custom version of C3 (slimming down the considerable size) would be appreciated!

It seems a much-requested feature but I could not find any working examples.

JamesMcMahon:
> these works for me, it will reduce d3 size by around 50%
> d3-array.js
> d3-brush.js
> d3-collection.js
> d3-color.js
> d3-dispatch.js
> d3-ease.js
> d3-format.js
> d3-interpolate.js
> d3-path.js
> d3-scale-chromatic.js
> d3-scale.js
> d3-selection.js
> d3-shape.js
> d3-time-format.js
> d3-timer.js
> d3-transition.js
> d3-zoom.js

How do you translate the above to config? Does anyone have an example?

PR #2686, opened 2019-08-19
correct DOMException when data name contains unicode

### LINKED ISSUE #2353: Uncaught DOMException when data label contains unicode
When a stacked area chart's group label contains unicode (`my\u007fapp`) this error occurs:

```
Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Element': '.c3-selected-circles-my\u007fapp' is not a valid selector
```

* **C3 version**: 0.5.3
* **D3 version**: 4.13.0
* **Browser**: Chrome 66.0.3359.81
* **OS**: Mac OS X

panthony:
This PR closes #2353 

In the end I did not introduce a new dependency.

I though about #2407 where size really matters and I did not feel like adding this polyfill for everyone.

This PR uses `CSS.escape` or fallback to previous method if not available, meaning the issue would still occurs for old browsers (ex: IE) but I think we'd better off explaining this and leaving the responsability to the user to embed a polyfill if needed.

### DIFF
--- a/src/class-utils.js
+++ b/src/class-utils.js
@@ -86,8 +86,13 @@ ChartInternal.prototype.classChartArc = function (d) {
     return CLASS.chartArc + this.classTarget(d.data.id);
 };
 ChartInternal.prototype.getTargetSelectorSuffix = function (targetId) {
-    return this.generateTargetClass(targetId)
-        .replace(/([?!@#$%^&*()_=+,.<>'":;\[\]\/|~`{}\\])/g, '\\$1');
+    const targetClass = this.generateTargetClass(targetId);
+    if (window.CSS && window.CSS.escape) {
+        return window.CSS.escape(targetClass);
+    }
+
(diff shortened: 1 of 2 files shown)


src/class-utils.js
```diff
@@ -86,8 +86,13 @@ ChartInternal.prototype.classChartArc = function (d) {
...
 };
 ChartInternal.prototype.getTargetSelectorSuffix = function (targetId) {
-    return this.generateTargetClass(targetId)
-        .replace(/([?!@#$%^&*()_=+,.<>'":;\[\]\/|~`{}\\])/g, '\\$1');
+    const targetClass = this.generateTargetClass(targetId);
+    if (window.CSS && window.CSS.escape) {
+        return window.CSS.escape(targetClass);
+    }
+
+    // fallback on imperfect method for old browsers (does not handles unicode)
```
kt3k: nice! 👍

PR #2687, opened 2019-08-19
more cache hits for getMaxTickWidth

panthony:
This PR:

- removes `withoutRecompute` and always return cached version if exists
- reuse `addToCache` and `getFromCache` instead of custom property
- add call to `resetCache` when `redraw` is called.

This way we have  more cache hits and the cache is dropped each time we want to redraw the chart (thinking of resize etc.) which should be, hopefully, sufficient.

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -309,40 +309,51 @@ Axis.prototype.textAnchorForY2AxisLabel = function textAnchorForY2AxisLabel() {
     var $$ = this.owner;
     return this.textAnchorForAxisLabel($$.config.axis_rotated, this.getY2AxisLabelPosition());
 };
-Axis.prototype.getMaxTickWidth = function getMaxTickWidth(id, withoutRecompute) {
+Axis.prototype.getMaxTickWidth = function getMaxTickWidth(id) {
     var $$ = this.owner,
         maxWidth = 0,
         targetsToShow, scale, axis, dummy, svg;
-    if (withoutRecompute && $$.currentMaxTickWidths[id]) {
-        return $$.currentMaxTickWidths[id];
+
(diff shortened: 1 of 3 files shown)


src/axis.js
```diff
@@ -309,40 +309,51 @@ Axis.prototype.textAnchorForY2AxisLabel = function textAnchorForY2AxisLabel() {
...
-                    var box = this.getBBox();
-                    if (maxWidth < box.width) {
-                        maxWidth = box.width;
-                    }
-                });
-                dummy.remove();
-            });
+
+    // bypass if SVG not rendered yet
+    if (!$$.svg) {
```
kt3k: nice early return 👍

src/core.js
```diff
@@ -533,6 +527,9 @@ ChartInternal.prototype.redraw = function(options, transitions) {
     durationForExit = withTransitionForExit ? duration : 0;
     durationForAxis = withTransitionForAxis ? duration : 0;
 
+    // reset caches when we are re-drawing the chart
+    $$.resetCache();
```
kt3k: I guess this resetting of cache might be inconvenient for some future use cases. How about resetting only some group of caches by the key?

```js
$$.resetCache('MaxTickWidth') // => reset cache which has the key /^\$MaxTickWidth/
```

panthony: I though about this when I tried to use this method for `legendItemTextBox` too where reseting this cache here would be unnecessary.

I think there will be some kind of tradeoff between ease of use (aka: Forget to reset a cache here & there) & performance.

I think there are 2 families of cached values:

- The ones that are impacted by an update of data (added, removed targets, etc.)
- The ones that are impacted by an update of sizing (resize, update of CSS....)

PR #2690, opened 2019-08-20
correct error when a pie starts empty

### LINKED ISSUE #2688: 0.7.5 release breaks pie chart
The following code works with 0.7.4:
```
c3.generate({
        bindto: `#mydiv`,
        size: { width: 150, height: 150 },
        padding: { right: 0, left: 0, top: 0, bottom: 0 }, // no effect but possible perf. optimization
        transition: {
          duration: 500,
        },
        interaction: {
          enabled: true,
        },
        pie: {
          label: {
            // By default c3 renders all values with floating point (10.0%)
            // This formatting function takes the number and renders it as: 10%
            format: (value) => {
              return `${value}%`;
            },
            /* threshold: 0.05 (5%) is the default which sticks out a bit but ok, 0.08 fits perfectly */
          },
        },
        data: {
          columns: [
            'r',
            'l',
            'd',
          ],
          type: 'pie',
          colors: {
            'r': 'red',
            'l': 'green',
            'd': 'blue',
          },
        },
        legend: {
          show: false,
        },
      });

      this.chart.load(
        {
          columns: [
            ['r', 75],
            ['l', 20],
            ['d', 5],
          ],
        },
      );
```
But with 0.7.5, it throws this error:
```
c3.js:7049 Uncaught TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at ChartInternal.getTotalDataSum (c3.js:7049)
    at ChartInternal.getRatio (c3.js:7519)
    at ChartInternal.textForArcLabel (c3.js:5562)
    at SVGTextElement.<anonymous> (d3-selection.js:585)
    at Selection.selection_each [as each] (d3-selection.js:365)
    at Selection.selection_text [as text] (d3-selection.js:592)
    at ChartInternal.redrawArc (c3.js:5880)
    at ChartInternal.redraw (c3.js:1816)
    at ChartInternal.initWithData (c3.js:1506)
```

panthony:
Closes #2688

### DIFF
--- a/spec/api.pie-spec.js
+++ b/spec/api.pie-spec.js
@@ -16,22 +16,58 @@ describe('c3 api pie', function () {
         }
     };
 
-    beforeAll(function (done) {
+    beforeEach(function (done) {
         chart = window.initChart(chart, args, done);
     });
 
-    it('can configure padAngle', function (done) {
-        expect(chart.pie.padAngle()).toBe(0.5);
+    describe('padAngle', function() {
 
-        const path = chart.internal.main.select('.c3-arc-data1').attr('d');
+        it('can configure padAngle', function (done) {
+            expect(chart.pie.padAngle()).toBe(0.5);
 
-        chart.pie.padAngle(0.2);
+            const path = chart.internal.main.select('.c3-arc-data1').attr('d');
 
-        setTimeout(function () {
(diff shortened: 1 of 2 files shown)


kt3k:
@panthony Thank you for the quick fix! 🏅

PR #2694, opened 2019-08-20
pass ratio to the data_labels_format callback when appropriate

panthony:
That one is a breaking change.

The `data_labels_format` callback now takes the `ratio` alongs with `value` and `index`.

The `ratio` is undefined if the chart is not normalized.

I also passed `$$.api` as this for the format callback in case it can be useful.

### DIFF
--- a/docs/reference.html.haml
+++ b/docs/reference.html.haml
@@ -1012,10 +1012,11 @@
           = partial :reference_item_link, locals: { id: 'data.labels.format' }
         %p Set formatter function for data labels.
         %br
-        %p The formatter function receives 4 arguments such as <span class="code">v</span>, <span class="code">id</span>, <span class="code">i</span>, <span class="code">j</span> and it must return a string that will be shown as the label. The arguments are:
+        %p The formatter function receives 5 arguments such as <span class="code">v</span>, <span class="code">ratio</span>, <span class="code">id</span>, <span class="code">i</span>, <span class="code">j</span> and it must return a string that will be shown as the label. The arguments are:
         %br
         %ul
           %li <span class="code">v</span> is the value of the data point where the label is shown.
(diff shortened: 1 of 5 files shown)


kt3k:
@panthony Isn't it possible to make this not a breaking change? How about making v have the shape of `{ value: number, ratio: number }` only when stack.normalize == true, or passing ratio as 5th argument?

panthony:
@kt3k I can pass it as 5th argument but it would mean that we'll have callbacks that takes:

> (value, ratio, id)

And others:

> (value, id, i, j, ratio)

Passing an object would still be a breaking change

kt3k:
@panthony 
> we'll have callbacks that takes:
(value, ratio, id)
And others:
(value, id, i, j, ratio)

Ah, that's bad...

Another option could be introducing new option data_labels_formatter (or something) which takes 5 arguments and keep data_labels_format take 4 arguments, though it looks a little tricky to implement and document...

kt3k:
Anyway we haven't introduced such explicit breaking changes recently except d3 v5 update, I'm going to talk with @masayuki0812 about the policy and how we update the version number if we do breaking changes.

panthony:
@kt3k Not sure if always possible but maybe we could pass directly the data point in all callbacks.

This way we have everything (id, index, value, ratio, any other attributes that may be added) and new attributes could be added in the future without API break.

To ease the upgrade we could introduce a boolean in the config to toggle the previous/new behavior.

Config that would eventually be dropped.

kt3k:
@panthony 

> This way we have everything (id, index, value, ratio, any other attributes that may be added) and new attributes could be added in the future without API break.

Sounds nice 👍 I prefer to switch to that API in some future.

> To ease the upgrade we could introduce a boolean in the config to toggle the previous/new behavior.

That's sounds reasonable. It would make the upgrade flow more graceful.

PR #2703, opened 2019-09-19
Rebasing region label work from PR 1853

dunderpate:
Hi @panthony, here's an updated patch for PR 1853.

The tests failed due to something seemingly unrelated (legend padding), and the same failure was also occurring on master. Apart from that, everything else seemed to work fine. Feel free to let me know if there's anything amiss and I'll be happy to fix it.

Thanks @panthony!


panthony:
@dunderpate Ideally we should document those new options in the [reference](https://github.com/c3js/c3/blob/master/docs/reference.html.haml#L1087).

And if this could be greenlighted by @kt3k that would be awesome 👍

dunderpate:
@panthony, thanks for the review. I'll take a stab at updating the reference documentation.

dunderpate:
Right on, ready for a review of the documentation when you have time @panthony. Thanks!

panthony:
@dunderpate Looks good! Thanks!

dunderpate:
Right on, thanks @panthony!

PR #2714, opened 2019-10-13
2289 - Add example of arc chart with fullCircle: true

ignoreintuition:


kt3k:
Thank you for the contribution, but it looks like you forgot adding `samples/chart_arc.html` to your commit.

ignoreintuition:
Yikes.  Sorry about that. Please check the latest PR which includes the additional file.  Also, please note, there is an issue with the fullCircle and with labels.  It appears to overlay the gauge on top of the labels.  Let me know if you want me to submit an issue and I can work it.

PR #2721, opened 2019-10-26
Update README.md

kruti-jain-stack:


kt3k:
Sorry. I don't get the point of this change.

Issue #2528, opened 2018-11-16
IE 11 legend toggle


vlarkovich:
**Steps to reproduce:**

1. Open any page with c3 chart in IE 11, for example [area chart](https://c3js.org/samples/chart_area.html) or [pie chart](https://c3js.org/samples/chart_pie.html)
2. Try to toggle any item on legend
3. See the chart with hidden item
![image](https://user-images.githubusercontent.com/43816231/48606008-722dcf00-e98f-11e8-9fff-f7de5f8f39dd.png)
 Or
![image](https://user-images.githubusercontent.com/43816231/48606094-b28d4d00-e98f-11e8-9f31-b195d3e806af.png)

**Expected result:** 
Data should not dissapear from chart.

Maybe it is because IE doesn't support `display: initial` css style and when api.show is called it doesn't remove `display: none` that was added in api.hide
https://github.com/c3js/c3/blob/9f546012615afa9b2c2fb85f36cbe7332cc374f7/src/api.show.js#L13

* **C3 version**: 0.6.8
* **D3 version**: 5.6.0
* **Browser**: IE 11
* **OS**: win10 x64

SovietAlex:
This was already mentioned before. 
[(https://github.com/c3js/c3/issues/2486)]

Just change
.style('display', 'initial', 'important') 
to
.style('display', 'block', 'important')

it works in all browsers after this

vlarkovich:
@SovietAlex, thank you, I've already fixed it locally, but I think that this bug has to be fixed in the main version of c3js, because this is a critical error.

snakada:
Hi, it might be a bit tricky, but I solved it by writing the following in CSS:

```css
.c3-target[style*="opacity: 1"] {
    display: block !important;
}
```

digEmAll:
I agree with @vlarkovich this should be included in the main version

GDFaber:
I'd like to make a PR ready for this, but it may take some time as this would be my first PR for this project and I have to set up a few things according to the contribution guidelines first.

GDFaber:
PR at #2741

PR #2741, opened 2020-01-16
Make legend toggle work in IE11

GDFaber:
- Replaced ```display: initial``` by ```display: block``` in Chart.prototype.show() to stop hidden data items from disappearing in IE11 when they should be shown again (as described in #2528).
- Built and tested the changes locally and run tests according to contributing guide.
- Did not add new tests; if I should do that here, please give me a pointer on how it is done in this case.

### DIFF
--- a/src/api.show.js
+++ b/src/api.show.js
@@ -10,7 +10,7 @@ Chart.prototype.show = function (targetIds, options) {
     targets = $$.svg.selectAll($$.selectorTargets(targetIds));
 
     targets.transition()
-        .style('display', 'initial', 'important')
+        .style('display', 'block', 'important')
         .style('opacity', 1, 'important')
         .call($$.endall, function () {
             targets.style('opacity', null).style('opacity', 1);
--- a/src/api.show.js
+++ b/src/api.show.js
@@ -1,4 +1,5 @@
 import { Chart } from './core';
+import { isIE } from './util';
 
 Chart.prototype.show = function (targetIds, options) {
     var $$ = this.internal, targets;
@@ -10,7 +11,7 @@ Chart.prototype.show = function (targetIds, options) {
     targets = $$.svg.selectAll($$.selectorTargets(targetIds));
 
     targets.transition()
-        .style('display', 'block', 'important')
+        .style('display', isIE() ? 'block' : 'initial', 'important')
         .style('opacity', 1, 'important')
         .call($$.endall, function () {
             targets.style('opacity', null).style('opacity', 1);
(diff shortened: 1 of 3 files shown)


kt3k:
@GDFaber Thank you for looking into this. The change looks almost good, but we'd like to minimize the effect of change. So can you add the conditional like the below instead of just changing to 'block'?

```
isIE ? 'block' : 'initial'
```

GDFaber:
Hi @kt3k,

sure, I'll add IE detection as stated above. I have a question about that:

I found a check for IE9 in clip.js/ChartInternal.prototype.getClipPath() as well. Should I add a detection function to util.js (as a part of this PR) and let getClipPath() use it (this would be a new PR then)? or is it ok to just check locally and leave clip.js as it is?

kt3k:
> Should I add a detection function to util.js (as a part of this PR) and let getClipPath() use it (this would be a new PR then)?

That sounds good to me.

GDFaber:
- Added functions for IE detection in utils.js
- Added basic unit tests for the IE detection functions

src/util.js
```diff
@@ -87,3 +87,38 @@ export var isWithinBox = function(point, box, sensitivity = 0) {
     return xStart < point[0] && point[0] < xEnd && yEnd < point[1] && point[1] < yStart;
 };
 
+/**
+ * Returns Internet Explorer version number (or false if no Internet Explorer used).
+ */
+export var getIEVersion = function() {
+    // https://stackoverflow.com/questions/19999388/check-if-user-is-using-ie
+    const agent = window.navigator.userAgent;
```
kt3k: How about taking `agent` as function paramter? If we do so, we can write the tests like the below:

```
expect(getIEVersion('Mozilla/5.0 (Windows NT 6.3; Trident/7.0; rv:11.0) like Gecko ')).toBe(11)
```

src/api.show.js
```diff
@@ -10,7 +11,7 @@ Chart.prototype.show = function (targetIds, options) {
     targets = $$.svg.selectAll($$.selectorTargets(targetIds));
 
     targets.transition()
-        .style('display', 'initial', 'important')
+        .style('display', isIE() ? 'block' : 'initial', 'important')
```
kt3k: nice!

src/util.js
GDFaber: Sure, let's do this. I'm going to modify getIEVersion() to take an optional parameter so usage is easy and can be tested as well.

Issue #2747, opened 2020-01-31
TypeError: c3.generate is not a function


lasseeee:
* **C3 version**: 0.7.12
* **D3 version**: 5.15.0

I just upgraded from 

C3 version: 0.7.11
D3 version: 5.15.0

to

C3 version: 0.7.12
D3 version: 5.15.0

And all charts that were working are now broken with the error:

`TypeError: c3.generate is not a function`

I'm loading the charts like this at the end of the file:

```
    $(function() {
        var chart = c3.generate({

        });
    });
```

kt3k:
How do you load c3.js? Do you use webpack?

lasseeee:
Yes. It's loaded using Webpack (Laravel Mix) with `window.c3 = require('c3');`

jakubsobel:
I have the same issue after the latest release. I'm using webpack 4.41.5 and latest changes in c3 exports broke the dynamic imports for me. The workaround is to use `c3.default.generate` or revert to older version of the c3.

In 0.7.11 this works fine:
```
import('c3')
  .then(c3 => c3.generate({...}))
```

In 0.7.12 I need to use:
```
import('c3')
  .then(c3 => c3.default.generate({...}))
```

kt3k:
https://github.com/c3js/c3/pull/2740 seems to have affected webpack build. Webpack uses `module` field instead of `main` if `module` is present in package.json.

PR #2748, opened 2020-02-03
Fix esm support

kt3k:
This PR tries to fix https://github.com/c3js/c3/issues/2747

- Webpack seems first trying to look up the entrypoint from `module` field instead of `main` field if it's present.

rollup users are still able to do `import c3 from 'c3/c3.esm'`


kt3k:
Probably what we should do is `export { generate, version, chart }` instead of `export default c3` in src/index.js. I'll update this PR soon.

PR #2790, opened 2020-06-01
hide the previous tooltip in case of null value

bounteous:
I am developing a bar graph in which it is necessary to differentiate the values of the "x" axis that have the value 0 or that simply there is no value for that position.

Since c3js does not support the use of "undefined" values, I am using a "null" to represent that there is no value on the "x" axis.

The problem comes when you have the mouse over a bar with a value and when you move the mouse over a bar that has no value, the tooltip of the previous bar remains visible.

The behavior that c3js currently has when rendering tooltips is somewhat visually confusing, as the position marked by the mouse is not clear.


Eus32:
It could be helpful in my project too.

Issue #2793, opened 2020-06-15
Point selection by dragging and unselect by clicking does not work in one condition


PT10:
* **C3 version**: 0.7.15 (latest)
* **D3 version**:5.16.0
* **Browser**:Chrome/Safari
* **OS**:Mac OS 10.14.5

There is a definite issue with the data point selection support. My data series has some selectable and some non selectable data points. I select the data points with a mouse drag to select few points. Now if I try to unselect them by clicking on them then they don't get unselected. However if I stick to the same method for unselect i.e. to drag the mouse again then the points get unselected. 

Also note that the behaviour is observed only when the data series has a combination of selectable and non selected points. If the entire series is selectable then the issue is not observed.

On deeper analysis I found out that actually the unselect event is getting triggered and the point is no more in the selected list but the circle which surrounds the point stays there. Hence it is an issue with the select circle. 

Here is the Stackblitz demo and the code.  [c3-chart-error](https://stackblitz.com/edit/c3-chart-example-seujfh?file=app%2Fapp.component.ts)

Here all points above 5 are selectable. Please follow the below steps to reproduce the issue.

1. Mouse drag/drop on points above value 5 to select them. (one or more)
2. Click on the selected point. The circle doesn't go away.

Thanks

PR #2795, opened 2020-06-15
Removing pre filter on drag and applying it while points traversal

PT10:
Fix for the issue #2793 

Because of the pre filter the index for the point was coming incorrect in mouse drag handler. Removed the filter and applied the check for isselectable inside the each loop.


PT10:
Also realised that the points selected using drag were not getting unselected with chart.unselect API if the series has a combination of selectable and non selectable points. That problem also will get fixed with this change.

